PR #20, opened 2026-03-30
Fix callback URL to include /api/automation prefix

malhotra5:
## Problem

Completion callbacks from SDK running inside sandboxes were not being registered properly. The callback URL passed when dispatching automation runs was missing the `/api/automation` prefix that's added by nginx middleware in production.

**Before:**
```
{base_url}/v1/runs/{run_id}/complete
```

**After:**
```
{base_url}/api/automation/v1/runs/{run_id}/complete
```

Without this prefix, the SDK's POST to the completion endpoint would fail to reach the automation service's router, causing:
- Completion callbacks to fail
- Runs to remain in RUNNING status until the watchdog eventually marks them as failed

### `automation/constants.py`
- Added `API_PATH_PREFIX = "/api/automation"` constant with documentation explaining its purpose

### `automation/dispatcher.py`
- Updated callback URL construction to include the `/api/automation` prefix
- Updated module docstring to reflect the correct endpoint path
- Imported the new `API_PATH_PREFIX` constant

### `tests/test_dispatcher.py`
Added `TestCallbackUrl` class with 3 unit tests:
- `test_callback_url_includes_api_automation_prefix` - Verifies the prefix is included
- `test_callback_url_with_custom_base_url` - Verifies custom base URL is used correctly  
- `test_callback_url_strips_trailing_slash` - Verifies no double slashes when base_url has trailing slash

## Testing

All new tests pass:
```bash
uv run pytest tests/test_dispatcher.py::TestCallbackUrl -v
# 3 passed
```

Tests are designed to run without Docker by mocking external dependencies.

### DIFF
--- a/automation/constants.py
+++ b/automation/constants.py
@@ -3,6 +3,12 @@
 from datetime import timedelta
 
 
+# API path prefix (added by middleware in production)
+# When the automation service runs as a sidecar, nginx routes requests through
+# /api/automation/* and strips the prefix. However, for callback URLs, the
+# sandbox SDK needs the full path including this prefix.
+API_PATH_PREFIX = "/api/automation"
+
 # Maximum allowed timeout for automation runs (10 minutes)
 # Used for:
 # - Validation in API requests (timeout field)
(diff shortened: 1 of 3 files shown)


malhotra5:
Actually this doesn't seems to be the root of the problem

Issue #23, opened 2026-03-31
Support automation presets for common use cases (prompt, plugins)


xingyaoww:
## Summary

Right now, the automation system works by uploading a full SDK script for every scheduled run. For the 80% use case where a user just wants to schedule a prompt (optionally with plugins), this feels heavyweight — we're re-uploading essentially the same boilerplate script each time. We should explore creating "presets" — predefined templates for common automation patterns.

## Motivation

Most users just want to schedule a prompt, possibly with plugins. Having them upload a full SDK script every time introduces unnecessary complexity and fragility. Presets simplify the user experience and avoid exposing the raw SDK script upload capability before we're ready to officially support it.

## Proposal

Create predefined templates for common automation patterns like "run a prompt" or "run a prompt with plugins." This could take the form of:

1. A higher-level API endpoint (potentially in SaaS) that accepts just a prompt string and plugin config, and generates the SDK script behind the scenes
2. A set of verified template scripts that read configuration (prompt, plugins) from environment variables, so we can reuse the same script across runs

The full SDK script upload remains available as a power-user escape hatch for advanced use cases.

## Acceptance Criteria

- [ ] Users can create a scheduled automation by providing just a prompt (without uploading a script)
- [ ] Users can optionally specify plugins alongside the prompt
- [ ] The system generates the appropriate SDK script behind the scenes
- [ ] Full SDK script upload is still supported for advanced use cases

PR #25, opened 2026-03-31
feat: add prompt-based automation creation endpoint

malhotra5:
## Summary

Add `POST /v1/from-prompt` endpoint that allows users to create automations by simply providing a prompt instead of manually creating and uploading a tarball.

### New Files
- **`automation/prompt_router.py`** - FastAPI router with:
  - `CreatePromptAutomationRequest` schema (name, prompt, trigger, timeout)
  - `_generate_tarball(prompt)` function to create tarball with SDK code
  - `POST /v1/from-prompt` endpoint

- **`automation/templates/`** directory:
  - `sdk_main.py` - SDK boilerplate that:
    - Fetches LLM config via `workspace.get_llm()`
    - Fetches secrets via `workspace.get_secrets()`
    - Fetches MCP config via `workspace.get_mcp_config()`
    - Creates `LLMSummarizingCondenser` for efficient context management
    - Creates `Agent` with LLM, tools, MCP config, and condenser
    - Executes the user's prompt via a `Conversation`
  - `setup.sh` - Installs SDK packages from PyPI (released versions)

- **`tests/test_prompt_router.py`** - Unit tests for tarball generation

### Modified Files
- **`automation/app.py`** - Register prompt_router
- **`pyproject.toml`** - Exclude `automation/templates` from ruff and pyright
- **`AGENTS.md`** - Documentation for the new feature

## How It Works

1. User sends `POST /v1/from-prompt` with:
   ```json
   {
     "name": "My Automation",
     "prompt": "Create a file called hello.txt with 'Hello World' inside",
     "trigger": {"type": "cron", "schedule": "0 9 * * 1", "timezone": "UTC"},
     "timeout": 300
   }
   ```

2. Service generates a tarball containing:
   - `main.py` - SDK code that reads `prompt.txt` and executes it
   - `prompt.txt` - The user's prompt
   - `setup.sh` - SDK installation

3. Uploads tarball to storage and creates both `TarballUpload` and `Automation` records

4. Returns the automation response with `tarball_path` pointing to `oh-internal://uploads/{uuid}`

## Testing

```bash
# Unit tests (tarball generation)
uv run pytest tests/test_prompt_router.py -v
```

All 4 unit tests pass. Integration tests require Docker and are appropriately skipped in CI without Docker.

---

*This PR was created by an AI assistant (OpenHands) on behalf of the user.*

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/cac56638-2749-43e7-8835-4aacb1e2c652)

Related to #23


automation/prompt_router.py
```diff
@@ -0,0 +1,185 @@
...
+        description=f"Auto-generated from prompt: {body.prompt[:100]}...",
+        status=UploadStatus.UPLOADING,
+        storage_path=storage_path,
+    )
+    session.add(upload)
+    await session.flush()
+
+    # Upload to storage (synchronous operation)
+    try:
+        file_store.write(path=storage_path, contents=tarball_content)
```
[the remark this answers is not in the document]
malhotra5: Fixed in 7f3677c. Changed from synchronous `write()` to async `write_stream()` using an async iterator helper `_bytes_to_async_iter()` to avoid blocking the event loop.

_This response was generated by an AI assistant (OpenHands)._

automation/prompt_router.py
```diff
@@ -0,0 +1,185 @@
...
+        name=body.name,
+        trigger=body.trigger.model_dump(),
+        tarball_path=tarball_path,
+        setup_script_path="setup.sh",
+        entrypoint="python main.py",
+        timeout=body.timeout,
+    )
+    session.add(automation)
+    await session.flush()
+    await session.refresh(automation)
```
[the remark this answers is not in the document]
malhotra5: Fixed in 7f3677c. Added try/except around automation creation with cleanup logic that calls `file_store.delete(storage_path)` if the automation record creation fails.

_This response was generated by an AI assistant (OpenHands)._

automation/prompt_router.py
```diff
@@ -0,0 +1,185 @@
...
+        tar.addfile(prompt_info, io.BytesIO(prompt_bytes))
+
+        # Add setup.sh from template
+        setup_sh_path = TEMPLATES_DIR / "setup.sh"
+        setup_sh_content = setup_sh_path.read_text()
+        setup_sh_bytes = setup_sh_content.encode("utf-8")
+        setup_sh_info = tarfile.TarInfo(name="setup.sh")
+        setup_sh_info.size = len(setup_sh_bytes)
+        setup_sh_info.mode = 0o755  # Make executable
+        tar.addfile(setup_sh_info, io.BytesIO(setup_sh_bytes))
```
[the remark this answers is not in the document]
malhotra5: Fixed in 7f3677c. Extracted `_add_file_to_tar()` helper function that reduces the tarball generation from 22 lines to 3 lines:

```python
_add_file_to_tar(tar, "main.py", templates["main.py"])
_add_file_to_tar(tar, "prompt.txt", prompt)
_add_file_to_tar(tar, "setup.sh", templates["setup.sh"], mode=0o755)
```

_This response was generated by an AI assistant (OpenHands)._

tests/test_prompt_router.py
```diff
@@ -0,0 +1,323 @@
...
+        assert data["trigger"]["schedule"] == "0 9 * * 1"
+        assert data["entrypoint"] == "python main.py"
+        assert data["setup_script_path"] == "setup.sh"
+        assert data["tarball_path"].startswith("oh-internal://uploads/")
+        assert data["enabled"] is True
+        assert "id" in data
+        assert data["user_id"] == str(TEST_USER_ID)
+
+        # Verify file store was called
+        mock_file_store.write.assert_called_once()
```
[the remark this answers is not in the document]
malhotra5: Fixed in 7f3677c. Improved test to:
1. Capture the actual tarball content from the mock
2. Verify the tarball contains `main.py`, `prompt.txt`, and `setup.sh`
3. Verify the prompt content matches exactly what was sent

_This response was generated by an AI assistant (OpenHands)._

automation/prompt_router.py
```diff
@@ -0,0 +1,185 @@
...
+
+    Returns:
+        bytes: The tarball content as bytes
+    """
+    tarball_buffer = io.BytesIO()
+
+    with tarfile.open(fileobj=tarball_buffer, mode="w:gz") as tar:
+        # Add main.py from template
+        main_py_path = TEMPLATES_DIR / "sdk_main.py"
+        main_py_content = main_py_path.read_text()
```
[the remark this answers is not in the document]
malhotra5: Fixed in 7f3677c. Added `_load_templates()` function with module-level `_TEMPLATE_CACHE` that loads templates on first access and returns cached values on subsequent calls.

_This response was generated by an AI assistant (OpenHands)._

pyproject.toml
```diff
@@ -55,6 +55,7 @@ dev = [
 [tool.ruff]
 target-version = "py312"
 line-length = 88
+extend-exclude = ["automation/templates"]
```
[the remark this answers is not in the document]
malhotra5: Acknowledged. The templates are excluded because they run inside the sandbox with SDK code patterns. I will consider adding syntax-checking in CI as a follow-up if needed.

_This response was generated by an AI assistant (OpenHands)._

automation/prompt_router.py
```diff
@@ -0,0 +1,185 @@
...
+    upload_id = uuid.uuid4()
+    storage_path = _build_storage_path(user.org_id, user.user_id, upload_id)
+
+    # Create upload record
+    upload = TarballUpload(
+        id=upload_id,
+        user_id=user.user_id,
+        org_id=user.org_id,
+        name=f"prompt-automation-{body.name[:50]}",
+        description=f"Auto-generated from prompt: {body.prompt[:100]}...",
```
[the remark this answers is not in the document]
malhotra5: Fixed in 7f3677c. Added `_safe_truncate()` helper function that properly encodes to UTF-8, slices at byte boundary, then decodes with `errors="ignore"` to handle multi-byte character splitting:

```python
def _safe_truncate(text: str, max_bytes: int) -> str:
    encoded = text.encode("utf-8")[:max_bytes]
    return encoded.decode("utf-8", errors="ignore")
```

_This response was generated by an AI assistant (OpenHands)._

automation/templates/__init__.py
xingyaoww, line 1: I wonder if we should put it as `templates/prompt_trigger`?

AGENTS.md
```diff
@@ -124,3 +124,40 @@ The SDK's `OpenHandsCloudWorkspace(local_agent_server_mode=True)` reads `SANDBOX
 - **Engine**: SQLAlchemy async with asyncpg; supports direct PostgreSQL (`AUTOMATION_DB_HOST`, `AUTOMATION_DB_PORT`, etc.) or GCP Cloud SQL connector (`AUTOMATION_GCP_DB_INSTANCE`)
 - **Migrations**: Alembic in `migrations/` directory
 - **Locking patterns**: `FOR UPDATE SKIP LOCKED` in scheduler/dispatcher polling, optimistic `UPDATE WHERE status=X` for callback/watchdog
+
+## Prompt-Based Automation Creation
+
+The `/v1/from-prompt` endpoint allows creating automations by simply providing a prompt, without manually creating and uploading a tarball.
```
xingyaoww: I wonder if we should consider name it `/template/prompt`, later we can extend it to maybe `/template/plugins` (but then we inevitably need to update automation code when plugin definition changes in the SDK.. 😓

malhotra5: @openhands make this change 

make sure that the route is placed at `v1/template/prompt` instead 

also create another subdirectory level `automation/templates/prompt` for the actual files containing the sdk code and setup script in `automation/templates/` and move the files under

malhotra5:
Done! I've made the following changes:

1. **Route changed**: `/v1/from-prompt` → `/v1/template/prompt`
2. **Templates moved**: `automation/templates/sdk_main.py` and `automation/templates/setup.sh` → `automation/templates/prompt/` subdirectory
3. **Updated files**:
   - `automation/prompt_router.py` - Updated router prefix and templates path
   - `automation/app.py` - Updated comment
   - `tests/test_prompt_router.py` - Updated endpoint paths
   - `AGENTS.md` - Updated documentation

All unit tests pass (4/4).

Commit: fc36bf9

_This comment was created by an AI assistant (OpenHands) on behalf of the user._

automation/templates/__init__.py
```diff
@@ -0,0 +1 @@
+"""Templates for prompt-based automation generation."""
```
xingyaoww: i think we need to change the doc string

automation/app.py
```diff
@@ -14,6 +14,7 @@
 from automation.db import create_engine, create_session_factory
 from automation.dispatcher import dispatcher_loop
 from automation.logger import setup_all_loggers
+from automation.prompt_router import router as prompt_router
```
xingyaoww: Shall we consider rename it as "preset_router" since it means we will create a few preset/template that user can readily used by plugging in argument instead of SDK scripts?

@OpenHands do some research and give some suggestions here. Do not modify this PR in any way

xingyaoww: Great question! Here's my analysis of the naming options, considering the current PR, the broader vision in issue #23, and internal consistency.

## Current naming landscape in this PR

| Artifact | Current name |
|---|---|
| Python module | `prompt_router.py` |
| Import alias | `prompt_router` |
| URL prefix | `/v1/template` |
| Endpoint | `POST /v1/template/prompt` |
| Router tags | `["Templates"]` |
| Templates dir | `automation/templates/prompt/` |

Notably, the URL and tags already use **"template"**, while only the Python module uses **"prompt"**.

## Option 1: `preset_router` (reviewer suggestion)

**Pros:**
- Aligns directly with issue #23's language: _"We should explore creating \"presets\" — predefined templates for common automation patterns."_
- Implies ready-to-use configurations where users plug in arguments
- Clearly distinguishes from the full SDK-script-upload path

**Cons:**
- Creates a terminology split: URL says `/template`, module says `preset` — you'd want to rename the URL prefix to `/v1/preset` and tags to `["Presets"]` for consistency
- "Preset" is less common in API/framework conventions (most use "template")

## Option 2: `template_router`

**Pros:**
- Already consistent with the URL prefix (`/v1/template`), tags (`Templates`), and directory (`automation/templates/`)
- Natural fit for the pattern: templates are parameterized (prompt text, trigger config) and generate a full automation
- Well-understood API convention — many platforms use "templates" for this concept (GitHub repo templates, Terraform modules, CloudFormation templates)
- Scales naturally: `POST /v1/template/prompt`, `POST /v1/template/prompt-with-plugins`, etc.

**Cons:**
- Could be confused with the literal Jinja/file templates in `automation/templates/` (though context usually disambiguates)

## Option 3: Keep `prompt_router` (status quo)

**Pros:**
- Accurately describes what this specific router does *right now*
- YAGNI — rename when a second preset is actually added

**Cons:**
- Once a second endpoint is added (e.g., plugins), the module name stops making sense
- Already inconsistent with the URL prefix and tags which use "template"

## Recommendation

I'd lean toward **`template_router`** as the strongest option:

1. **Zero URL/tag changes needed** — the prefix is already `/v1/template` and tags are `["Templates"]`
2. **Minimal rename** — just the Python file and import alias
3. **Scales naturally** as new presets are added under the same router
4. **Industry convention** — "template" is the standard term for parameterized blueprints in API design

If the team feels strongly about "preset" (matching issue #23's wording), that works too — but then I'd recommend also renaming the URL prefix to `/v1/preset`, tags to `["Presets"]`, and the directory to `automation/presets/` for full consistency.

Either way, `prompt_router` is the weakest option long-term since it's already inconsistent with the URL scheme and won't generalize.

_This analysis was generated by an AI assistant (OpenHands) on behalf of the user._

malhotra5: The SDK code behavior from this endpoint is set in stone for the prompt endpoint (e.g always include llm, secrets, and mcp config)

For this reason I'd prefer preset over template

automation/templates/__init__.py
malhotra5: Fixed in dade51e. Renamed `automation/templates/` to `automation/presets/` with subdirectory `prompt/` for the prompt-based preset files.

_This response was generated by an AI assistant (OpenHands)._

automation/templates/__init__.py
malhotra5: Fixed in dade51e. Updated the docstring in `automation/presets/__init__.py` to describe presets as ready-to-use automation configurations and list available presets.

_This response was generated by an AI assistant (OpenHands)._

automation/app.py
malhotra5: Fixed in dade51e. Renamed to `preset_router` with full consistency:

- Module: `automation/preset_router.py`
- Import: `from automation.preset_router import router as preset_router`
- URL prefix: `/v1/preset`
- Endpoint: `POST /v1/preset/prompt`
- Router tags: `["Presets"]`
- Directory: `automation/presets/`

This naming better reflects the concept of ready-to-use automation configurations and scales naturally for future presets.

_This response was generated by an AI assistant (OpenHands)._

pyproject.toml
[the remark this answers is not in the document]
malhotra5: Great! If the PR is worth merging and the solution is pragmatic please make sure to approve the PR

automation/presets/prompt/sdk_main.py
```diff
@@ -0,0 +1,163 @@
+"""Prompt-based automation script — runs inside an OpenHands Cloud sandbox.
```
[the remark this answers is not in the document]
malhotra5: Fixed in c882db1. Added `TestPresetFileSyntax` class with two tests:

1. `test_prompt_preset_sdk_main_syntax` - Uses `compile()` to verify `sdk_main.py` has valid Python syntax
2. `test_prompt_preset_setup_sh_exists` - Verifies `setup.sh` exists and contains expected content

This provides a safety net since preset files are excluded from linting, catching syntax errors before they break user automations.

_This response was generated by an AI assistant (OpenHands)._

automation/presets/prompt/sdk_main.py
```diff
@@ -0,0 +1,163 @@
...
+        keep_first=2,
+    )
+    print("  LLMSummarizingCondenser configured (max_size=100, keep_first=2)")
+
+    # Create agent with tools, MCP config, and condenser
+    print("\n=== AGENT ===")
+    tools = [
+        Tool(name=TerminalTool.name),
+        Tool(name=FileEditorTool.name),
+    ]
```
xingyaoww: I wonder if we can use get_default_tools / get_default_agent from openhands.tools here

automation/preset_router.py
```diff
@@ -0,0 +1,226 @@
...
+from automation.storage import FileStore, get_file_store
+from automation.utils.tarball_validation import build_internal_url
+
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter(prefix="/v1/preset", tags=["Presets"])
+
+# Preset files directory for prompt-based automations
+PRESETS_DIR = Path(__file__).parent / "presets" / "prompt"
```
xingyaoww: Shall we call it `PROMPT_PRESETS_DIR` since we will likely include other stuff later

automation/preset_router.py
```diff
@@ -0,0 +1,226 @@
...
+        id=upload_id,
+        user_id=user.user_id,
+        org_id=user.org_id,
+        name=f"prompt-automation-{_safe_truncate(body.name, 50)}",
+        description=f"Auto-generated from prompt: {truncated_prompt}...",
+        status=UploadStatus.UPLOADING,
+        storage_path=storage_path,
+    )
+    session.add(upload)
+    await session.flush()
```
xingyaoww, line 166: I wonder if we can pre-upload the version of "presets" code, and have the prompt be exposed in other ways (eg env vars)... This way we don't have to repeatly upload the same tarball over and over again, we can at least save some space.

Although this might require us to change something in our arch - might not be worth it.

@OpenHands can you do a thorough analysis and provide some suggestions here

xingyaoww: Great question! Here's a thorough analysis of the optimization and the tradeoffs involved.

## Current Architecture (Per-Automation Tarball)

Every `POST /v1/preset/prompt` call:
1. Generates a tarball containing `main.py` (5.8KB), `setup.sh` (241B), and `prompt.txt` (user's prompt)
2. Uploads it to GCS → creates `TarballUpload` record
3. Creates `Automation` pointing to `oh-internal://uploads/{uuid}`

Every **run** of that automation:
1. Dispatcher downloads the tarball from GCS (~2.4KB compressed)
2. Uploads it to the sandbox
3. Sandbox extracts, runs setup.sh, then `python main.py` which reads `prompt.txt`

The boilerplate (`main.py` + `setup.sh`) is **identical** across all prompt-based automations. Only `prompt.txt` varies.

## Option A: Shared Preset Tarball + Prompt via Env Var

**How it would work:**
- Upload `main.py` + `setup.sh` **once** as a shared "preset tarball" (per version)
- Store the prompt in a new column on the `Automation` model (e.g., `metadata JSONB`)
- Dispatcher injects `AUTOMATION_PROMPT` env var when dispatching
- `main.py` reads `os.environ["AUTOMATION_PROMPT"]` instead of `prompt.txt`
- All prompt automations share the same `tarball_path`

**Required changes:**
1. `Automation` model: Add `metadata` JSONB column + DB migration
2. `dispatcher.py`: Read `metadata.prompt` and inject as `AUTOMATION_PROMPT` env var
3. `sdk_main.py`: Read prompt from env var instead of file
4. `preset_router.py`: Store prompt in metadata, reference shared tarball
5. New: Shared tarball lifecycle management (upload on startup/deploy, versioning)

**Pros:**
- ✅ Storage savings: 1 tarball for N automations instead of N tarballs
- ✅ Reduces GCS reads per run (all runs fetch same object → highly cacheable)
- ✅ Updating the preset code is simpler (update one tarball, all automations benefit)

**Cons:**
- ⚠️ **Env var size limits**: Prompts can be up to 50KB. The dispatcher exports env vars via shell commands (`export KEY='...'`), which are sent as HTTP POST bodies to the agent-server bash API. While there's no kernel `ARG_MAX` issue (it's not `execve()`), 50KB shell-quoted strings in a single command could be fragile.
- ⚠️ **Cross-cutting dispatcher changes**: The dispatcher currently treats all automations uniformly. Adding preset-aware logic (check metadata, inject prompt env var) breaks this clean separation.
- ⚠️ **Shared tarball lifecycle**: Who uploads it? When? How do we version it? What happens during a rolling deploy when the preset code changes?
- ⚠️ **DB migration**: Needs a new column and migration to production.

## Option B: Content-Addressed Deduplication

**How it would work:**
- Hash the boilerplate files (main.py + setup.sh) to get a content key
- Before uploading, check if a tarball with the same code hash already exists for this org
- Reuse existing tarball when code matches, only create new when prompt differs

**Assessment:** Doesn't help — every prompt IS different, so every tarball is unique regardless.

## Option C: Hybrid — Shared Tarball + Prompt as Separate File Upload

**How it would work:**
- Shared preset tarball contains only `main.py` + `setup.sh`
- Prompt stored in a separate small file in GCS (or in DB metadata)
- Dispatcher downloads both, combines them into the sandbox
- Or: dispatcher uploads tarball, then writes prompt to a file via a second bash command

**Pros:** Avoids env var size limits, still shares boilerplate
**Cons:** Even more complex — two-step dispatch, more error handling

## Size Analysis

| Scenario | Storage per automation | GCS read per run |
|---|---|---|
| Current (embedded) | ~2.4KB | ~2.4KB |
| Shared tarball | ~0 (shared) + prompt in DB | ~1.8KB (shared) |

At scale: **10,000 automations × 2.4KB = 24MB** of GCS storage. Even 100,000 would be 240MB. GCS costs are ~$0.02/GB/month, so this is essentially free.

## Recommendation

**For this PR: Keep the current approach.** The per-automation tarballs are tiny (~2.4KB), storage costs are negligible, and the architecture is simple and self-contained. Each automation is a complete, independent unit.

**For a follow-up (if/when needed):** The shared tarball + env var approach (Option A) becomes worthwhile when:
1. We add more presets with **larger** boilerplate code
2. We need to **update preset code** across all existing automations (versioning story)
3. The automation table grows to hundreds of thousands of entries

If we do pursue it later, I'd suggest:
- Add a `metadata JSONB` column to `Automation` (useful for all presets, not just prompt)
- Use a `preset_type` field to let the dispatcher know it should inject preset-specific env vars
- Keep prompt size limit reasonable for env vars (~32KB) and fall back to file-based for larger prompts
- Upload the shared tarball as part of service startup (idempotent, keyed by content hash)

The architecture change is modest but touches the dispatcher's core path, so it's better as a dedicated follow-up than bolted onto this PR.

_This analysis was generated by an AI assistant (OpenHands)._

automation/presets/prompt/sdk_main.py
malhotra5: I think the problem for default agent is that you can't easily pass in mcp config, but we can probably just model copy that's more maintainable

@openhands read the docs here for sdk code: https://docs.openhands.dev/sdk

use default agent, use default tools, make sure to include the mcp config using model_copy on the resulting default agent though

malhotra5: Fixed in f660ef8. Updated `sdk_main.py` to use `get_default_agent()` from `openhands.tools` with `cli_mode=True` (to disable browser tools), then use `model_copy(update={"mcp_config": mcp_config})` to add the MCP configuration.

This approach:
- Uses the SDK's recommended default tools and condenser configuration
- Is more maintainable as it automatically inherits any future updates to the default setup
- Properly passes MCP config using Pydantic's `model_copy()` method

_This response was generated by an AI assistant (OpenHands)._

automation/preset_router.py
malhotra5: Good point! The variable is currently named `PRESETS_DIR` which points to `presets/prompt/` specifically. Since we will likely add other presets (e.g., plugins), I think keeping it as `PRESETS_DIR` (pointing to `presets/prompt/`) for this file is fine since this router only handles the prompt preset.

Alternatively, if we add more presets that share this router, we could have a parent `PRESETS_ROOT_DIR` pointing to `presets/` and then construct paths like `PRESETS_ROOT_DIR / "prompt"`. But for now, the current naming seems appropriate.

Let me know if you'd prefer a different approach!

_This response was generated by an AI assistant (OpenHands)._

xingyaoww: ^This router is actually "Preset" router, so it is expected to handle more than just prompt

xingyaoww (approved):
LGTM except a nit https://github.com/OpenHands/automation/pull/25#discussion_r3023553698

automation/preset_router.py
malhotra5: Fixed in 40349b5. Renamed to be more specific for future extensibility:

- `PRESETS_DIR` → `PROMPT_PRESET_DIR`
- `_PRESET_CACHE` → `_PROMPT_PRESET_CACHE`
- `_load_preset_files()` → `_load_prompt_preset_files()`

_This response was generated by an AI assistant (OpenHands)._

PR #35, opened 2026-04-03
Rewrite automation service to use Temporal

malhotra5:
## Summary

This PR rewrites the automation service to use [Temporal](https://temporal.io/) for durable workflow execution, replacing the legacy polling-based scheduler/dispatcher/watchdog system.

### Removed (Legacy Code)
- `scheduler.py` - Cron polling loop → replaced by **Temporal Schedules**
- `dispatcher.py` - Pending run polling → replaced by **Temporal Workflows**
- `watchdog.py` - Stale run detection → replaced by **Temporal timeouts**
- `execution.py` - Sandbox lifecycle → moved to **Temporal Activities**
- `utils/run.py`, `utils/cron.py`, `utils/api_key.py` - Legacy utilities

### Added (Temporal Implementation)
| File | Purpose |
|------|---------|
| `temporal/types.py` | Data classes for workflow inputs/outputs |
| `temporal/activities.py` | 6 activities for sandbox operations |
| `temporal/workflows.py` | AutomationWorkflow with crash-proof execution |
| `temporal/schedules.py` | Temporal Schedule management for cron jobs |
| `temporal/client.py` | Temporal client factory |
| `temporal/worker.py` | Worker setup |

### Updated
- `app.py` - FastAPI app now runs Temporal worker as background task
- `router.py` - API routes use Temporal for workflow execution
- `config.py` - Added Temporal connection settings

## Benefits of Temporal

1. **Fault Tolerance**: Workflows automatically resume after crashes
2. **Built-in Timeouts**: No need for watchdog polling
3. **Schedules**: First-class cron scheduling with ScheduleSpec
4. **Activity Retries**: Automatic retries with configurable policies
5. **Visibility**: Rich UI for monitoring workflow execution

## Tests Added

| Test File | Description | Count |
|-----------|-------------|-------|
| `test_types.py` | Data class tests | 12 |
| `test_schedules.py` | Schedule CRUD tests | 11 |
| `test_activities.py` | Activity tests with ActivityEnvironment | 12 |
| `test_workflows.py` | Workflow tests (skipped by default) | 4 |

## Running Tests

```bash
# Unit tests (fast, no external deps)
uv run pytest tests/temporal/test_types.py tests/temporal/test_schedules.py -v

# Activity tests
uv run pytest tests/temporal/test_activities.py -v

# All temporal tests (excluding workflow tests)
uv run pytest tests/temporal/ -v
```

## Deployment Notes

Requires Temporal server. For Kubernetes deployment, Temporal can be deployed as pods. Configuration via environment variables:

```
AUTOMATION_TEMPORAL_HOST=temporal.default.svc.cluster.local
AUTOMATION_TEMPORAL_PORT=7233
AUTOMATION_TEMPORAL_NAMESPACE=default
AUTOMATION_TEMPORAL_TASK_QUEUE=automations
```

---
*This PR was created by an AI assistant (OpenHands) on behalf of the user.*


automation/temporal/workflows.py
```diff
@@ -0,0 +1,329 @@
...
+    API_KEY_RETRY_POLICY,
+    SANDBOX_RETRY_POLICY,
+    TARBALL_RETRY_POLICY,
+    EXECUTION_RETRY_POLICY,
+    CLEANUP_RETRY_POLICY,
+) = _get_retry_policies()
+
+
+@workflow.defn
+class AutomationWorkflow:
```
xingyaoww, line 115: Maybe call it AutomationExecutionWorkflow?

I'd imagine we will have some other workflows

automation/temporal/activities.py
```diff
@@ -0,0 +1,472 @@
...
+    from sqlalchemy import select
+
+    from automation.db import create_engine, create_session_factory
+    from automation.models import TarballUpload
+    from automation.storage import get_file_store
+
+    logger.info("Downloading internal tarball: upload_id=%s", input.upload_id)
+
+    settings = get_settings()
+    engine_result = await create_engine(settings)
```
jlav, line 212: You can share a database connection pool across all your activities. Creating an engine here creates a new pool with new connections every time.

automation/temporal/workflows.py
```diff
@@ -0,0 +1,329 @@
...
+            },
+        )
+
+        settings = get_settings()
+        sandbox_info: SandboxInfo | None = None
+        api_key: str | None = None
+
+        try:
+            # 1. Get per-user API key
+            api_key = await workflow.execute_activity(
```
jlav, line 147: Activity output is stored in plain text in the temporal database, and is visible in the Temporal UI by default. Not a great place to store credentials.

Also, only return information from an activity that is guaranteed to be static until the end of the workflow execution. If these API keys are short lived and the workflow becomes long-lived, it's bound to fail permanently once the API Key expires.

Depending on the lifetime here, sometimes it's better to just re-retrieve info inside of individual activities. If it's long lived, you can look into data converters to encrypt the info: https://docs.temporal.io/default-custom-data-converters#default-data-converter

automation/temporal/workflows.py
```diff
@@ -0,0 +1,329 @@
...
+                    start_to_close_timeout=timedelta(minutes=5),
+                    retry_policy=TARBALL_RETRY_POLICY,
+                )
+            else:
+                # Internal tarball - download from storage, upload to sandbox
+                upload_id = parse_internal_upload_id(tarball_path)
+                if upload_id is None:
+                    raise ValueError(f"Invalid tarball_path: {tarball_path}")
+
+                tarball_data = await workflow.execute_activity(
```
jlav, line 196: The largest factor in workflow throughput is the size of the activity payloads. These tarballs are going to be stored directly workflow history, inside of postgres. Postgres isn't a good place to store large binary data.

Factor this out to store tarballs somewhere externally and have the activity retrieve the contents directly. 

`EXTERNAL_MAX_FILESIZE` is 100MB right now. The max per-event payload size defaults to 4MB in temporal, and the max history is 50MB with warnings emitted at 10MB. Increasing those defaults cripples performance, and right now any file larger than ~4MB would just fail altogether.

tests/temporal/test_workflows.py
jlav, line 1: Strongly recommend just biting the bullet and adding replay tests as early as possible. They'll catch non-determinism errors that you might introduce by modifying the workflow after it's already been deployed. When you make code changes to a workflow, if there are any active workflows when you deploy the change they are likely to fail if they aren't done in a backwards compatible way.

You can read more here on replay tests: https://docs.temporal.io/develop/python/testing-suite#replay

jlav (commented):
Strongly recommend adding a docker compose setup that has the temporal dev server set up (and any other dependencies like postgres) for ease of local development: https://learn.temporal.io/getting_started/python/dev_environment/

Having the temporal UI available locally for debugging is super useful.

PR #37, opened 2026-04-08
feat: serve app under /api/automation base path

jlav:
## Why

We currently rely on a Traefik `StripPrefix` middleware in our Helm charts to route `/api/automation` traffic to this service. This couples the application to a specific ingress implementation and makes it harder for consumers using other ingress controllers (e.g. nginx, AWS ALB) to deploy. By having the app handle the `/api/automation` prefix directly, we can remove that middleware and simplify ingress configuration.

## Notable details

- The `root_path` config property has been removed — it was only needed when a reverse proxy stripped the prefix before forwarding. Since the app now owns the path, `root_path` would cause double-prefixing in Swagger UI.
- The `resolved_base_url` localhost fallback now includes `/api/automation` so that sandbox callback URLs are correct in local dev without needing `AUTOMATION_BASE_URL` set.

### DIFF
--- a/automation/app.py
+++ b/automation/app.py
@@ -134,10 +134,7 @@ def _build_cors_origins() -> list[str]:
 
 def _create_app() -> FastAPI:
     """Create and configure the FastAPI application."""
-    settings = get_settings()
-    base_path = settings.base_path
-
-    application = FastAPI(
+    return FastAPI(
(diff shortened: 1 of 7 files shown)


malhotra5 (approved):
LGTM

PR #54, opened 2026-04-20
chore: Add frontend npm build

raymyers:
Adding ci to frontend, no dockerfile yet.

This is the "Configure CI pipeline" section of PLTF-1240

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 ---
 # .github/workflows/precommit.yml
-name: Pre-commit checks
+name: ci
 
 on:
     push:
@@ -9,7 +9,7 @@ on:
         branches: ['**']
 
 jobs:
-    pre-commit:
+    backend:
         runs-on: ubuntu-24.04
 
         steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,6 @@
 ---
 # .github/workflows/precommit.yml
```
hieptl, line 2: **suggestion (blocking):** The file has been renamed, so I believe this line should be updated to `# .github/workflows/ci.yml`. Thank you! 🙏

.github/workflows/ci.yml
```diff
@@ -29,3 +29,28 @@ jobs:
...
             - name: Run pre-commit (all files)
               run: uv run pre-commit run --all-files --show-diff-on-failure
+
+    frontend:
+        runs-on: ubuntu-24.04
+        defaults:
+            run:
+                working-directory: frontend
+        steps:
+            - uses: actions/checkout@v4
```
hieptl: **question (non-blocking):** I noticed that we’re using `actions/checkout@v5` in the backend and `actions/checkout@v4` in the frontend. Could you help clarify why v4 is being used for the frontend? Thank you! 🙏

.github/workflows/ci.yml
```diff
@@ -29,3 +29,28 @@ jobs:
...
+            - name: npm dependencies
+              run: npm ci
+
+            - name: lint
+              run: npm run lint
+
+            - name: test
+              run: npm run test
+
+            - name: build
```
hieptl: **suggestion (non-blocking):** I noticed that the backend uses Title case for step names (e.g., “Checkout code,” “Install uv”), while the frontend uses lowercase (e.g., “lint,” “test,” “build”). Could we align these conventions for consistency across both? Thank you! 🙏

raymyers: Fair enough, done.

.github/workflows/ci.yml
raymyers: Aligning on 5, thanks.

hieptl (approved):
Thank you! 🙏

PR #70, opened 2026-04-25
fix: serve OpenAPI docs at base path for correct proxy routing

jpshackelford:
## Problem

The OpenAPI documentation endpoints (`/openapi.json`, `/docs`, `/redoc`) are not accessible when the automation service is deployed behind a reverse proxy at `/api/automation`. 

This regression was introduced in PR #37 when `root_path` was removed.

## Solution

Set `docs_url`, `openapi_url`, and `redoc_url` explicitly to include the base path (`/api/automation`), making the OpenAPI spec and documentation UIs accessible at:

- `/api/automation/openapi.json`
- `/api/automation/docs` (Swagger UI)
- `/api/automation/redoc`

## Why this approach?

This is consistent with PR #37's direction of having the app **own all its paths directly**, rather than relying on `root_path`. The original fix in PR #30 used `root_path`, but PR #37 removed it because:

> "The `root_path` config property has been removed — it was only needed when a reverse proxy stripped the prefix before forwarding. Since the app now owns the path, `root_path` would cause double-prefixing in Swagger UI."

By explicitly setting the docs URLs, we maintain Joe's architectural direction while fixing the accessibility issue.

## Testing

Added `tests/test_openapi.py` with tests verifying all three endpoints are accessible.

---

@jpshackelford can click here to [continue refining the PR](https://app.all-hands.dev/conversations/f5e9894c-52a2-4578-8393-7c431ad04dd0)

### DIFF
--- a/automation/app.py
+++ b/automation/app.py
@@ -139,6 +139,9 @@ def _build_cors_origins() -> list[str]:
 
 def _create_app() -> FastAPI:
     """Create and configure the FastAPI application."""
+    # Serve OpenAPI docs under the base path so they're accessible when the app
+    # is mounted at /api/automation (e.g., /api/automation/docs).
+    base_path = get_settings().base_path
     return FastAPI(
         title="OpenHands Automations Service",
         description=(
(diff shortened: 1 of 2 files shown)


jpshackelford:
## ✅ Verified in Preview Deployment

Confirmed the fix works in the preview environment:

- **OpenAPI JSON**: https://au-pr-70.staging.all-hands.dev/api/automation/openapi.json
- **Swagger UI**: https://au-pr-70.staging.all-hands.dev/api/automation/docs

Both endpoints are now accessible as expected.

aivong-openhands (approved):

PR #144, opened 2026-05-28
fix(presets): pin venv to Python >=3.12 in setup.sh

DevinVinson:
uv venv with no --python defaults to the first system interpreter, which on macOS is CommandLineTools Python 3.9. That fails the subsequent openhands-sdk install since the SDK requires Python >=3.12.

This fixes https://github.com/OpenHands/agent-canvas/issues/2157 for me when I load locally:
```
OH_AUTOMATION_REPO=https://github.com/DevinVinson/automation \
OH_AUTOMATION_GIT_REF=dv/setup-venv-python-312 \
npm run dev
```


malhotra5 (approved):
LGTM, thanks!

PR #159, opened 2026-06-04
PLTF-2895: restore missing indexes on automations and automation_runs

### LINKED ISSUE #157: perf: create missing indexes on automation_runs and automations tables
## Problem

All non-primary-key indexes on `automation_runs` and `automations` tables are **missing in production** despite being defined in Alembic migration `001_initial_schema.py`. Alembic reports version `006` (all migrations applied), but the indexes were never actually created.

### Evidence (production CloudSQL, 2026-06-03)

| Table | seq_scan | seq_tup_read | idx_scan | n_live_tup |
|-------|----------|-------------|----------|------------|
| `automation_runs` | 4,360,000+ | **61,700,000,000** (61.7B) | 0 (pkey only) | — |
| `automations` | 5,380,000+ | **484,000,000** (484M) | 0 (pkey only) | — |

In a 10-second sample: ~173 seq scans on `automation_runs`, ~151 on `automations`.

### Missing Indexes (defined in migration 001 but absent)

**`automation_runs`:**
- `ix_automation_runs_automation_id`
- `ix_automation_runs_status`
- `ix_automation_runs_pending` (partial: status=PENDING)
- `ix_automation_runs_timeout_at`

**`automations`:**
- `ix_automations_user_id`
- `ix_automations_org_id`
- `ix_automations_enabled`
- `ix_automations_deleted_at`
- `ix_automations_last_polled_at`

### Critical Query Patterns Affected

1. **Scheduler poll**: `WHERE enabled = true AND next_run_at <= now AND last_polled_at < threshold`
2. **Dispatcher poll**: `WHERE status = 'PENDING'`
3. **Watchdog scan**: `WHERE status = 'RUNNING' AND timeout_at < now`
4. **Listing queries**: `WHERE user_id = ? AND deleted_at IS NULL` / `WH4. **Listing queries**: `WHERE user_id = ? AND deleted_at IS NULL` / `WH4. **Listins using `CREATE INDEX CONCURRENTLY` (safe for live deployment)
2. Investigate why migration 001 didn't create the indexes — possible partial failure during initial deployment, or the tables were created manually without the indexes
3. Add a smoke test or startup check that verifies expected indexes exist

### Context

These tables share the same CloudSQL instance as the main OpenHands app.

---

aivong-openhands:
Closes #157

## Problem
Every non-PK index on `automations` and `automation_runs` is **missing in production** — confirmed against the prod `automations` DB, only `automations_pkey` and `automation_runs_pkey` exist — even though they're declared on the models and in migration `001`, and alembic reports the latest revision.

This makes the dispatcher poll (`automation_runs WHERE status = 'PENDING' ORDER BY created_at ASC … FOR UPDATE SKIP LOCKED`) do a full table scan on every poll — it's currently among the top queries by total execution time on the shared CloudSQL instance.

## Root cause
The indexes were almost certainly added to migration `001` **after** it had already been applied in prod. Editing an already-applied migration never re-runs it, so already-migrated databases never got the indexes (fresh deployments that ran the corrected `001` do have them). A new migration is the fix.

## Fix
Migration `007` recreates all 9 indexes exactly as declared in the models / migration `001`, with `if_not_exists=True` so it's a **no-op** where they already exist:

**automations:** `user_id`, `org_id`, `enabled`, `deleted_at`, `last_polled_at`
**automation_runs:** `automation_id`, `status`, partial `(created_at) WHERE status='PENDING'`, `timeout_at`

No model changes — the models already declare these.

### Plain `CREATE INDEX`, not `CONCURRENTLY`
#157 suggests `CONCURRENTLY`, but this migration harness runs on **pg8000 inside a transaction holding an advisory lock**, where alembic's `autocommit_block` (required for `CONCURRENTLY`) fails — the same incompatibility recently hit on an OpenHands enterprise migration. Both tables are small in prod (`automation_runs` ~52K rows / 16 MB, `automations` ~200 rows), so each build is sub-second and the brief write-lock is acceptable.

## Follow-ups (not in this PR)
- Add the startup/smoke check from #157 point 3 to assert expected indexes exist, so this can't silently recur.
- Other models also declare `index=True` columns (e.g. `TarballUpload`, `CustomWebhook`) — worth verifying those indexes exist in prod too.

## Verification plan
On deploy I'll confirm the indexes exist and that the dispatcher poll switches from Seq Scan to an index/partial-index scan, and capture the before/after seq-scan rate (same method used for the `event_callback` fix).

### DIFF
--- a/migrations/versions/007_restore_missing_indexes.py
+++ b/migrations/versions/007_restore_missing_indexes.py
@@ -1,0 +1,97 @@
+"""Restore missing indexes on automations and automation_runs
+
+The non-primary-key indexes for these tables are declared on the models and in
+migration 001, but are absent in production: alembic reports the latest revision
+yet only the PK indexes exist. The most likely cause is that the indexes were
+added to migration 001 after it had already been applied, so they were never
+created in already-migrated databases (editing an applied migration does not
+re-run it). This migration recreates them idempotently.
+
+Effect: removes full table scans for the scheduler/dispatcher/watchdog polls and
+the listing queries, notably the dispatcher poll
+``WHERE status = 'PENDING' ORDER BY created_at`` (covered by the partial
+ix_automation_runs_pending) — one of the heaviest queries on the shared CloudSQL
+instance.
+
+Plain CREATE INDEX (not CONCURRENTLY): this migration harness runs on pg8000
+inside a transaction holding an advisory lock, where alembic's autocommit_block
+(required for CREATE INDEX CONCURRENTLY) fails. Both tables are small in prod
+(automation_runs ~52K rows / 16MB, automations ~200 rows), so each build is
+sub-second and the brief write lock is acceptable. if_not_exists keeps this a
+no-op on databases that already have the indexes (e.g. fresh deployments).
+
+Revision ID: 007
+Revises: 006
+Create Date: 2026-06-04
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+from sqlalchemy import text
+
+revision: str = "007"
+down_revision: Union[str, None] = "006"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # automations
+    op.create_index(
+        "ix_automations_user_id", "automations", ["user_id"], if_not_exists=True
+    )
+    op.create_index(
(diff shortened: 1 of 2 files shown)


aivong-openhands:
@openhands fix failing backend check

aivong-openhands:
Might not be needed so setting to draft

PR #163, opened 2026-06-06
fix: support external Postgres SSL mode

ak684:
## Summary
- Add `AUTOMATION_DB_SSL_MODE` support for automation service direct Postgres connections.
- Propagate SSL mode through asyncpg runtime connections and pg8000 Alembic migrations.
- Support `prefer`, `require`, and `disable` with focused unit coverage.
  - Successfully validated `sslmode=require` on a real Replicated OHE instance using an external Postgres endpoint on port `5433`; app login and a test conversation passed.
  
## Validation
- `uv run pytest tests/test_config.py tests/test_db.py`
- `git diff --check`

### DIFF
--- a/migrations/env.py
+++ b/migrations/env.py
@@ -23,6 +23,7 @@ import os
 from alembic import context
 from sqlalchemy import create_engine, text
 
+from openhands.automation.db import _build_pg8000_connect_args
 from openhands.automation.models import Base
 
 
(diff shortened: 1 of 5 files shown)


ak684:
Agreed with the env-variable consistency issue. I pushed `eac5478` to make `ServiceSettings.db_ssl_mode` use the same fallback chain as migrations: `AUTOMATION_DB_SSL_MODE` -> `DB_SSL_MODE` -> `PGSSLMODE`, with tests covering fallback and precedence.

PR #179, opened 2026-06-10
fix: make preset bootstraps shell-neutral across platforms

### LINKED ISSUE #178: [fix]: make preset automations cross-platform beyond setup.sh
### **Problem**
automation#172 and #177 fixed important Windows blockers, but preset automations still depend on setup.sh and POSIX shell assumptions. Native Windows support is therefore still not truly out of the box.

### **Scope**
- Remove or supplement setup.sh with a cross-platform bootstrap path for prompt/plugin presets
- Ensure the preset create/run path works on Windows, macOS, and Linux
- Keep the Windows entrypoint and timezone improvements from #172/#177 covered by tests

### **Goal**
Preset automations should not require Git Bash or WSL on Windows.

### **Out of scope**
New automation product/tool abstractions and broader CI/process changes.
☝️ We contributors need to discuss this long term solution. But this bandage should work OK for now 🩹 .

And, the human user read & amended this too.

jamiechicago312:
## Summary
- replace preset setup.sh dependence with cross-platform bootstrap.py launchers
- route setup_script_path=None runs through a Python runner instead of POSIX shell exports/bash
- add Windows-safe regression coverage for preset tarballs and execution paths

## Testing
- uv run pytest tests/test_execution.py tests/test_preset_router.py tests/test_ab_testing_integration.py -q
- uv run ruff check openhands/automation/execution.py openhands/automation/preset_router.py openhands/automation/dispatcher.py tests/test_execution.py tests/test_preset_router.py tests/test_ab_testing_integration.py

Closes #178


jamiechicago312:
I did have OH write this but, FR I just ran into this when recording a demo, it sucks

--jamie
---
@malhotra5 I just hit this in a real Agent Canvas on Windows flow, and I want to emphasize that this PR feels critical for smooth Windows support.

Automations really need to work on the first try for Windows users in agent-canvas. Right now, a simple automation can still fail before the actual prompt even starts if any POSIX-only path or shell assumptions leak through the local automation bootstrap/dispatch path.

From the user side, that makes Windows automation support feel unreliable even when the actual automation prompt is fine. Getting this merged would meaningfully improve the out-of-box experience for Windows users and reduce a very sharp edge in local agent-canvas automations.

_This comment was posted by an AI agent (OpenHands) on behalf of Jamie Chicago._

openhands/automation/execution.py
```diff
@@ -188,6 +192,87 @@ async def _upload(
     resp.raise_for_status()
 
 
+def _get_python_launcher() -> str:
+    """Return a cross-platform Python launcher for inline runner commands."""
+    return "py -3" if os.name == "nt" else "python"
+
+
+def _build_python_runner_command(
```
malhotra5, line 213: Ah this is clever, although I think we don't assume that python command is available on the system (we lean on uv for the virtual env, and we can assume that uv exists)

Although, that's only for preset automations as well it may not be true for custom sdk automations

openhands/automation/constants.py
```diff
@@ -24,11 +28,10 @@
...
 
-# Path where tarballs are extracted inside the sandbox. This is:
-# - Written by the sandbox initialization script
-# - Read by the automation entrypoint
-# DO NOT CHANGE: Would break tarball extraction in running sandboxes.
-TARBALL_PATH = "/tmp/automation.tar.gz"
+# Path where tarballs are stored before extraction. Uses the platform temp
+# directory so the agent-server accepts the path on every OS (Windows
+# requires a drive-letter prefix for os.path.isabs() to return True).
+TARBALL_PATH = os.path.join(tempfile.gettempdir(), "automation.tar.gz")
```
malhotra5: Is this still being used anywhere? I think this is legacy, we create unique tmp dirs nowadays to avoid extracting multiple tarballs in the same location

jamiechicago312:
@openhands can you resolve the request in this comment https://github.com/OpenHands/automation/pull/179#pullrequestreview-4527230556
push to this pr please

openhands/automation/presets/plugin/bootstrap.py
```diff
@@ -0,0 +1,97 @@
+"""Cross-platform preset bootstrap for plugin automations.
```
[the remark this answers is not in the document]
jamiechicago312: Fixed in 9991fef: extracted a single shared `openhands/automation/presets/bootstrap.py` and deleted the duplicate `prompt/bootstrap.py` and `plugin/bootstrap.py`. Both `_load_prompt_preset_files()` and `_load_plugin_preset_files()` now reference `BOOTSTRAP_PATH` from the shared file. New preset types only need one bootstrap file going forward.

openhands/automation/presets/prompt/bootstrap.py
```diff
@@ -0,0 +1,97 @@
...
+packages, then re-execs the generated main.py with the venv interpreter.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import shutil
+import subprocess
+import sys
```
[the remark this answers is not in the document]
jamiechicago312: Fixed in 9991fef: removed `import sys` from the shared `presets/bootstrap.py`.

openhands/automation/presets/prompt/bootstrap.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def main() -> None:
+    uv_path = _require_uv()
+    sdk_version = _fetch_sdk_version()
+    print(f"[bootstrap] Creating virtual environment in {VENV_DIR}")
+    _run_checked(uv_path, "venv", str(VENV_DIR), "--python", PYTHON_REQUIREMENT, "--quiet")
+
+    print(f"[bootstrap] Installing OpenHands SDK packages at version {sdk_version}")
+    install_args = [uv_path, "pip", "install", "--quiet"]
```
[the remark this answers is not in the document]
jamiechicago312: Fixed in 9991fef: `uv pip install` now passes `--python str(_venv_python_path())` explicitly, so the venv target is correct regardless of CWD.

jamiechicago312:
Addressed all three issues from the review (commit 9991fef):

**[Critical] DRY Violation — resolved**
Extracted a single shared `openhands/automation/presets/bootstrap.py`. Both `prompt/bootstrap.py` and `plugin/bootstrap.py` are deleted (197 lines removed). `preset_router.py` now loads bootstrap content from `BOOTSTRAP_PATH = PRESETS_DIR / 'bootstrap.py'` for both preset types. Future presets will automatically share the same bootstrap with no duplication.

**[Suggestion] Unused `import sys` — resolved**
Removed from the shared bootstrap.

**[Suggestion] Implicit venv targeting — resolved**
`uv pip install` now passes `--python str(_venv_python_path())` explicitly, making the target venv unambiguous regardless of CWD.

Tests: 87 passed, 0 failed (`uv run pytest tests/test_preset_router.py tests/test_execution.py -q`). Ruff clean on all application files.

openhands/automation/execution.py
jamiechicago312: Interesting
I feel when the automation has been running it always is using python3 and then it fails because Windows wont accept that.

openhands/automation/constants.py
jamiechicago312: /tmp/ fails on windows.

jamiechicago312: @openhands can you address this comment from malhotra5?

openhands/automation/execution.py
jamiechicago312: @openhands can you address this request from malhotra5?

jamiechicago312: Fixed in 3827532.

Replaced the platform-specific `_get_python_launcher()` logic (`python` on POSIX / `py -3` on Windows) with `uv run python` on all platforms — since `uv` is guaranteed to be present in the sandbox, this avoids assuming a system-level Python installation is in `PATH`. The same change was applied to `PRESET_BOOTSTRAP_ENTRYPOINT` in `preset_router.py`, which now also uses `uv run python bootstrap.py` uniformly.

_This comment was posted by an AI agent (OpenHands) on behalf of the user._

openhands/automation/constants.py
jamiechicago312: Good catch — `TARBALL_PATH` is not a true protocol constant (unlike `WORK_DIR`, which the SDK must agree on). It is only an internal temp-file path used by `run_automation()` (the E2E test helper) and as a last-resort fallback in `execute_in_context()` when `run_id` is absent. The sandbox itself has no fixed expectation for this filename.

Fixed in 3bc6c9d: moved the path out of `constants.py` into `execution.py` as a private `_FALLBACK_TARBALL_PATH`, keeping `constants.py` limited to true protocol constants (`WORK_DIR`, `MODEL_PROFILE_PATTERN`). The `os`/`tempfile` imports that were added to `constants.py` only to support this value are removed as well.

_This comment was posted by an AI agent (OpenHands) on behalf of the user._

jamiechicago312:
Addresses comments and retried

https://github.com/user-attachments/assets/37e3fa1f-6092-4da0-824d-9befbcce9853

jamiechicago312:
I think everything has been addresses @malhotra5 tysm

PR #185, opened 2026-06-11
feat: grant permissive CORS to API-key requests

hieptl:
## Summary

Browser clients that authenticate with API keys can now call the automation API directly — cross-origin — without a server-side proxy hop. Cookie/session traffic is completely unaffected.

This unblocks the agent-server GUI's automations page, which currently tunnels every cloud call through the bundled agent-server's `POST /api/cloud-proxy` because this service rejects its CORS preflights.

## Problem

All cross-origin requests were answered by a single strict `CORSMiddleware` (allowlist from `AUTOMATION_CORS_ORIGINS`, fallback `openhands_api_base_url`, `allow_credentials=True`). The agent-server GUI on `http://localhost:3001` therefore failed preflight on `/api/automation/*`. The main cloud API solved the identical problem in OpenHands/OpenHands#14473; this service lagged behind, and the GUI singled it out with a `forceProxy` workaround.

## Changes

- **`openhands/automation/middleware.py` (new):** `ApiKeyAwareCORSMiddleware` — a CORS dispatcher wrapping two `CORSMiddleware` policies:
  - **Permissive** (`Access-Control-Allow-Origin: *`, credentials off) for requests authenticating via `Authorization: Bearer …` or `X-Session-API-Key`. Detected on preflights via whole-word matching of `Access-Control-Request-Headers` (so `x-my-authorization-token` cannot substring-match), and on actual requests via the headers themselves. The wildcard is safe because browsers cannot attach cookies when credentials are off.
  - **Strict** (existing allowlist via the unchanged `_build_cors_origins()`, credentials on) for cookie/anonymous traffic — identical to the previous behavior.
  - Deliberate differences from the OpenHands reference: no RFC 8628 device-flow path exemptions and no `X-Access-Token` detection (this service supports neither).
- **`openhands/automation/app.py`:** register `ApiKeyAwareCORSMiddleware` in place of the plain `CORSMiddleware`.

## Tests

New `tests/test_cors.py` (9 tests):
- **Middleware dispatch matrix** (isolated app): API-key preflights and actual requests from a foreign origin get wildcard CORS without credentials; credential-less preflights from a foreign origin are blocked; allowlisted-origin cookie traffic keeps strict CORS with credentials; lookalike-header regression guard.
- **Real-app integration:** `OPTIONS`/`GET /api/automation/health` with API-key headers from `http://localhost:3001` get `Access-Control-Allow-Origin: *` (previously the preflight was answered `400` with no allow-origin header).

Full suite: **782 passed**. `ruff` and `pyright` clean.

### DIFF
--- a/openhands/automation/app.py
+++ b/openhands/automation/app.py
@@ -8,7 +8,6 @@ from contextlib import asynccontextmanager
 from pathlib import Path
 
 from fastapi import FastAPI
-from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse
 from fastapi.staticfiles import StaticFiles
 from sqlalchemy import text
@@ -23,6 +22,7 @@ from openhands.automation.db import (
 from openhands.automation.dispatcher import dispatcher_loop
 from openhands.automation.event_router import router as event_router
 from openhands.automation.logger import setup_all_loggers
+from openhands.automation.middleware import ApiKeyAwareCORSMiddleware
 from openhands.automation.preset_router import router as preset_router
 from openhands.automation.router import router
 from openhands.automation.scheduler import scheduler_loop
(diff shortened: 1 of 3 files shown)


tofarr (approved):
🍰

PR #197, opened 2026-06-18
fix(watchdog): process each stale run in its own session to avoid lock contention

tofarr:
## Problem

Two slow queries were observed in production on the `automation_runs` table:

| Query | Avg time | Calls | Avg rows returned |
|---|---|---|---|
| `UPDATE … SET status, error_detail, completed_at WHERE id=? AND status=?` | 6.24s | 92 | 0.261 |
| `UPDATE … SET status, completed_at WHERE id=? AND status=?` | 19.1s | 10 | 0.5 |

Both filter by primary key (`id`), so the slowness is not an indexing problem — a PK lookup is nanoseconds. Multi-second execution on a single-row UPDATE is always lock contention.

## Root cause

`mark_stale_runs` in `watchdog.py` opened **one database session for the entire stale-run batch**:

```python
async with session_factory() as session:
    stale_runs = ...  # SELECT all stale runs

    for run in stale_runs:
        await _verify_and_mark_run(session, run, settings)
        # ↑ calls backend.verify_run() — slow external HTTP to sandbox
        # then executes UPDATE WHERE id=? AND status=RUNNING
        # → row lock acquired, NOT yet released

    if marked:
        await session.commit()   # ← locks released HERE, after the whole batch
```

Each `UPDATE` inside `_verify_and_mark_run` acquired a row-level exclusive lock that was held until the single `session.commit()` at the very end — across every subsequent `backend.verify_run()` call. When 10 stale runs were in a batch and each verification took ~2s, the first locked row was held for ~18s after its own UPDATE ran.

Callback endpoint UPDATEs (`WHERE id=? AND status=RUNNING`) on those same rows queued behind the watchdog's open transaction, producing the 6–19s waits. The < 1 avg rows returned confirms the optimistic lock was frequently lost: by the time the waiting UPDATE ran, the watchdog had already changed the status.

## Fix

Fetch only the stale run IDs in a short-lived read session that closes immediately, then process each run in its own dedicated session, committing right after `_verify_and_mark_run`. Row locks are released as soon as each run is processed rather than held for the duration of the batch.

```python
# 1. Short read session — closes before any per-run work
async with session_factory() as session:
    stale_run_ids = [...]

# 2. One session per run — lock released immediately after commit
for run_id in stale_run_ids:
    async with session_factory() as session:
        ...
        if await _verify_and_mark_run(session, run, settings):
            await session.commit()   # lock released here, not at end of batch
```

## Testing

All 7 existing watchdog unit tests pass. The optimistic locking behaviour (`AND status = RUNNING`) is fully preserved.

---
_This pull request was created by an AI agent (OpenHands) on behalf of the user._

### DIFF
--- a/openhands/automation/watchdog.py
+++ b/openhands/automation/watchdog.py
@@ -212,8 +212,7 @@ async def mark_stale_runs(
         # Fetch stale run IDs only — close this session before doing any
         # per-run work so we don't hold locks across slow verify calls.
         result = await session.execute(
-            select(AutomationRun.id)
-            .where(
+            select(AutomationRun.id).where(
                 AutomationRun.status == AutomationRunStatus.RUNNING,
                 AutomationRun.timeout_at.isnot(None),
                 AutomationRun.timeout_at < now,
--- a/openhands/automation/watchdog.py
+++ b/openhands/automation/watchdog.py
@@ -199,27 +199,41 @@ async def mark_stale_runs(
     Before marking as FAILED, attempts to verify the actual status by querying
     the sandbox. Uses optimistic locking so concurrent callbacks win.
 
+    Each run is processed in its own session so that row locks are released
+    immediately after commit rather than held for the duration of the batch.
+    This prevents lock contention with concurrent callback UPDATEs.
+
     Returns the number of runs marked with terminal status.
     """
     now = utcnow()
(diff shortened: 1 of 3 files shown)


malhotra5 (approved):
LGTM!

PR #223, opened 2026-06-30
fix: capture active LLM profile from the correct users/me field

hieptl:
### Issue

Resolves https://github.com/OpenHands/agent-canvas/issues/2080

### Summary
Automations should pin the LLM profile that was active when they were created so that changing the global active profile later never alters an existing automation. The machinery for this already exists (PR OpenHands/agent-canvas#2349) but was silently non-functional — a one-line field-name bug meant the active profile was never captured. This PR fixes that.

### Problem
- The GUI creates automations without specifying a `model`, expecting the backend to capture the user's active profile.
- Instead, every automation was persisted with `model = NULL` and executed against **whatever LLM profile was globally active at runtime** — so changing the active profile silently changed existing automations.

### Root cause
The service reads the active profile from `GET /api/v1/users/me`, which serializes it under `llm_profiles.active`. `_extract_active_model_profile_name()` read `llm_profiles.active_profile` — the key used by the *different* `/api/v1/settings/profiles` endpoint. The lookup therefore always returned `None`. An existing unit test hid the bug because its fixture mocked `/users/me` with the `/settings/profiles` shape.

### Changes
- **`openhands/automation/auth.py`** — read `profiles_payload.get("active")` instead of `"active_profile"`. One line; the existing capture → persist → execute chain is otherwise untouched.
- **`tests/test_auth.py`** — correct the stale fixture to the real `/users/me` shape (`active` key, `profiles` as a name-keyed dict), and add `test_authenticate_no_active_profile_when_none_selected` covering the "profiles present but none active → not pinned → runtime-default fallback" boundary.

No changes needed in the dispatcher, SDK presets, schema, ORM model, GUI, or OpenHands — those were already correct.

### Testing
- `uv run pytest tests/test_auth.py -q` → **47 passed** (corrected regression test + new boundary test).
- `uv run pytest -q` (full automation suite) → **1021 passed**, no regressions.
- The corrected `test_authenticate_extracts_model_profile_metadata` now fails if `auth.py` is reverted to `active_profile`, locking in the fix.

### Design decision — pre-existing automations (`model = NULL`)
Left as-is (no data migration). The profile active at their creation was never recorded, and the automation DB has no access to per-user active profiles, so a backfill has nothing correct to write; "today's active profile" would just re-introduce the global coupling we're removing. `NULL ⇒ runtime default` stays the documented, backward-compatible contract.


openhands/automation/auth.py
tofarr, line 495: I wonder if we should use pydantic here rather than custom logic based around dictionaries? (So it fails fast if there is an incompatible API change?)

tofarr (changes requested):
See my comment - LMK if you feel this increases the scope of this PR too much and should be a follow up PR.

openhands/automation/auth.py
hieptl: Hello @tofarr, I have updated the code based on your feedback. Thank you very much! 🙏

tofarr (approved):
🍰

PR #224, opened 2026-06-30
fix: support ACP agent_kind in preset automation scripts

### LINKED ISSUE #164: Automation preset scripts fail when agent server is configured with ACP agent (agent_kind: "acp")
When the agent server is configured to use an ACP agent (`agent_kind: "acp"`), cron automations created via the prompt/plugin presets fail immediately with:

```
LLMBadRequestError: litellm.BadRequestError: LLM Provider NOT provided.
Pass in the LLM provider you are trying to call. You passed model=acp
```

**Root cause:** The generated preset `main.py` unconditionally calls `get_default_agent(llm=llm, cli_mode=True)`, which creates a standard `Agent` and tries to route LLM calls through LiteLLM using `llm.model`. When the server is in ACP mode, the persisted `llm.model` is the sentinel string `"acp"` (used only for cost attribution), not a real LiteLLM model identifier.

**Repro steps:**
1. Configure agent server with `agent_kind: "acp"` (e.g. `devin acp`).
2. Create a prompt-based automation (no `model` override).
3. Wait for cron trigger or dispatch manually.
4. Run fails with the above error.

**Expected behavior:** The preset script should detect the server's `agent_kind` and either:
- Use `settings.create_agent()` for ACP (returns `ACPAgent` with subprocess command, args, env), or
- Continue with `get_default_agent()` for standard `openhands` mode.

**Actual behavior:** `sdk_main.py` (both prompt and plugin presets) always uses `get_default_agent(llm=llm, ...)` and ignores the configured agent type.

**Environment:**
- `openhands-automation` version: `1.0.0a6`
- `openhands-sdk` version: `1.26.0`

**Suggested fix:** In the preset templates, replace the hardcoded `get_default_agent()` call with something like:

```python
settings = workspace._fetch_agent_settings()
agent = settings.create_agent()
# ... then apply cli_mode / mcp_config / agent_context as needed
```

This leverages the existing discriminated union (`AgentSettingsConfig`) that already correctly returns `ACPAgent` or `Agent` based on `agent_kind`.

Hades32:
## Summary

Preset-based automations (prompt and plugin) failed immediately when the agent server was configured with an ACP agent (`agent_kind: "acp"`), because the generated `sdk_main.py` unconditionally called `get_default_agent(llm=llm, cli_mode=True)`. That constructs a standard `Agent` and routes LLM calls through LiteLLM using `llm.model` — but in ACP mode the persisted `llm.model` is the sentinel string `"acp"` (cost attribution only, not a real LiteLLM model), so the run crashed with:

```
LLMBadRequestError: litellm.BadRequestError: LLM Provider NOT provided.
Pass in the LLM provider you are trying to call. You passed model=acp
```

Fixes #164.

## Changes

- `openhands/automation/presets/prompt/sdk_main.py`
- `openhands/automation/presets/plugin/sdk_main.py`

In both preset templates, replaced the hardcoded `get_default_agent()` call with agent-kind detection:

1. Fetch agent settings via `workspace._fetch_agent_settings()` — the same SDK helper that `get_llm()` and `get_mcp_config()` already use internally.
2. If the settings are `ACPAgentSettings` → call `settings.create_agent()`, which returns an `ACPAgent` that delegates to a subprocess ACP server (no LiteLLM routing of the sentinel model).
3. Otherwise (`openhands` / legacy `llm`) → keep the existing `get_default_agent(llm=llm, cli_mode=True)` path, preserving current behavior.

The MCP config / `agent_context` overlay via `model_copy` is preserved. For ACP, `get_mcp_config()` returns `{}` (the `mcp_config` field only exists on `OpenHandsAgentSettings`), so the ACP agent's own `mcp_config` from `create_agent()` is not clobbered.

## Testing

- Added assertions to `test_generate_tarball_main_py_content` and `test_generate_plugin_tarball_main_py_content` verifying the ACP detection code (`ACPAgentSettings`, `_fetch_agent_settings`, `create_agent()`) is present in the generated `main.py`.
- All 84 tests in `tests/test_preset_router.py` pass.
- Verified both preset files parse as valid Python (`compile()` / `ast.parse`).


Hades32:
cc @malhotra5

Hades32:
or maybe @hieptl

Hades32:
@hieptl this is blocking me using automations with a pure ACP-based setup

smolpaws:
Merged `main` into this branch and resolved the conflicts so the fix stays up to date with upstream.

**What changed in the resolution:**
- Both `presets/prompt/sdk_main.py` and `presets/plugin/sdk_main.py` now keep this PR's `_resolve_agent()` helper (ACP detection via `_fetch_agent_settings()` → `create_agent()`), while preserving the `finish_tool_response_schema=TaskOutcome` wiring that landed in `main` via #334.
- `_resolve_agent()` gained a `finish_tool_response_schema` parameter so the non-ACP path forwards it to `get_default_agent()`, exactly as `main` now does. The ACP path still calls `create_agent()` (which takes no such parameter).

**Why:** `main` had independently changed the agent construction in both templates after this PR branched, so a plain merge conflicted. The resolution intends to change nothing about either side's behavior — only to keep both features working together.

All 98 tests in `tests/test_preset_router.py` pass (41 pre-existing skips).

PR #228, opened 2026-07-04
fix: bump SDK to 1.31.0

neubig:
## Summary

- Bump `openhands-sdk` and `openhands-workspace` from 1.29.3 to 1.31.0.
- Regenerate `uv.lock`.

## Why

`openhands-automation` runs generated automation scripts with the SDK version exposed by its `/sdk-version` endpoint. The 1.29.3 SDK rejects current agent-server event payloads that include `parent_id`, so automation must consume the same current SDK event models instead of requiring agent-server to hide new event fields.

## Validation

- `uv lock --check`
- `uv run python - <<'PY' ...` confirmed installed `openhands-sdk` and `openhands-workspace` both resolve to `1.31.0`.


enyst (approved):

hieptl (approved):
Thank you! 🙏

PR #257, opened 2026-07-24
fix: normalize MCP config shapes in automation presets

malhotra5:
## Summary

- Normalize MCP config returned by workspace helpers in both prompt and plugin automation presets.
- Accept both SDK-native MCP server maps and wrapped `{ "mcpServers": ... }` config before passing the config to the agent.
- Update preset tarball tests to assert the normalization path and guard against the old `mcp_config.get("mcpServers")` check.

## Testing

- `uv run pytest tests/test_preset_router.py -k "sdk_main_syntax or main_py_content" -q`

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/85862088-741e-4f4d-885d-98141a658266)


VascoSch92 (approved):
One clarification: with the currently pinned SDK 1.37.0, both workspace implementations already return dict[str, MCPServer]. The old mcp_config.get("mcpServers") check only affected logging; the native map was still passed to the agent via model_copy(update=...). 

So in the supported runtime this fixes the misleading “no MCP servers configured” message rather than missing MCP functionality. The wrapped shape was used before SDK 1.31.2.

Non-blocking suggestions:
- Replace the source-string assertions—including one against the docstring—with behavioral tests covering native, wrapped, empty, and None inputs.
- Consider either exporting coerce_mcp_config publicly or handling ImportError defensively, since the current submodule import creates an SDK ≥1.31.2 requirement.

Overall, the implementation is safe with the current exact SDK pin.

malhotra5:
Addressed the review feedback in ec17faf: replaced the brittle MCP source-string assertions with behavioral coverage for native, wrapped, empty, and None inputs, and added a defensive fallback for environments where the SDK coercer import is unavailable.\n\n_This comment was created by an AI agent (OpenHands) on behalf of the user._

PR #267, opened 2026-07-27
fix(security): keep injected secrets out of commands

simonrosenberg:
HUMAN:

Make sure secrets are not logged.

---

AGENT:

## Why

Automation environment values were embedded directly in the bash command sent to the agent server. Runtime timeout and failure logging could therefore persist session keys, API keys, and integration credentials as command text.

## Summary

- Upload injected environment values through the authenticated file API instead of embedding them in the command JSON.
- Disable xtrace, restrict the file to mode 600, source it, and remove it immediately with an EXIT cleanup fallback.
- Validate environment variable names and cover quoted values, command redaction, file loading, and cleanup.

## How to Test

- uv run pytest tests/ -v --ignore=tests/integration
  - 1067 passed on the production change.
- uv run pytest tests/test_execution.py -q
  - 32 passed on the final commit, including a real bash subprocess that loads a quoted value and verifies the file is removed.
- uv run pre-commit run --files openhands/automation/execution.py tests/test_execution.py
  - All hooks passed.

## Rollout

Deploy together with the SDK runtime-log hardening PR for defense in depth: this removes credentials from generated commands, while the SDK change removes raw command and output fields from logs.

No credential values or production log payloads are included in this PR.


aivong-openhands:
## Mutation-testing the new suite

I ran mutation testing against `TestPrivateEnvironmentInjection` to see which parts of the change the tests actually pin down. The suite is not inert — three control mutants get caught:

| Control mutant | Result |
|---|---|
| `_serialize_env_vars` stops shell-quoting the value | ❌ caught by `test_env_file_is_loaded_and_removed` |
| Revert `execute_in_context` to inline `export …` in the command | ❌ caught by `test_execute_command_omits_env_values` |
| Delete the start-failure `finally` cleanup block | ❌ caught by `test_cleanup_failure_preserves_permanent_start_error` |

So the *stated* fix — secrets out of the command text — is well covered. What isn't covered is the machinery defending it. Four mutants survive with a fully green `tests/test_execution.py` (34 passed):

| Surviving mutant | Result |
|---|---|
| Delete `env_path = None` after a successful `_start_bash` | ✅ 34 passed |
| Delete `trap {cleanup} EXIT` from `_env_command_prefix` | ✅ 34 passed |
| Delete the `_ENV_VAR_NAME_RE` check in `_serialize_env_vars` | ✅ 34 passed |
| Delete the in-chain `rm -f` (leaving the trap as sole cleanup) | ✅ 34 passed |

Below is how I'd kill each. All four are additions to the existing class; no new harness needed.

---

### 1. `env_path = None` — highest consequence

`test_execute_command_omits_env_values` doesn't patch `_bash`, so if the `finally` block did fire it would run against the `AsyncMock` client and the test would still pass. Nothing asserts cleanup is *skipped* on success.

That matters because the failure mode is severe: if that reset is ever lost, the service `rm`s the env file out from under a command it just started, the in-sandbox chain dies at `.`, and every automation with env vars stops running — while dispatch still reports `success=True`.

```python
@pytest.mark.asyncio
@patch("openhands.automation.execution._bash", new_callable=AsyncMock)
@patch("openhands.automation.execution._start_bash", new_callable=AsyncMock)
@patch("openhands.automation.execution._upload", new_callable=AsyncMock)
async def test_successful_start_leaves_env_cleanup_to_the_command(
    self, mock_upload, mock_start_bash, mock_bash
):
    mock_start_bash.return_value = "cmd-123"

    result = await execute_in_context(
        client=AsyncMock(),
        agent_url="https://agent.example.com",
        session_key="session-key",
        entrypoint="python main.py",
        tarball_source=b"fake tarball bytes",
        work_dir=DEFAULT_WORK_DIR,
        env_vars={"PRIVATE_VALUE": "secret"},
    )

    assert result.success is True
    mock_bash.assert_not_awaited()  # the in-sandbox chain owns the removal
```

### 2. The `EXIT` trap

`test_env_file_is_loaded_and_removed` only walks the happy path, where the explicit `rm -f` does the removal — the trap is never the thing that cleans up, so deleting it changes nothing observable. The trap is the entire fallback for "chain aborted before `rm`", which is reachable today: `chmod`, `.`, and `rm` are all in an `&&` chain.

Break the chain at `.` by making the sourced file unparseable — then only the trap can have removed it:

```python
def test_env_file_removed_when_sourcing_fails(self, tmp_path):
    env_path = tmp_path / "private.env"
    env_path.write_bytes(b"export BAD=(\n")  # syntax error -> `.` fails

    result = subprocess.run(
        ["bash", "-c", f"{_env_command_prefix(str(env_path))}echo unreachable"],
        capture_output=True,
        text=True,
    )

    assert result.returncode != 0
    assert "unreachable" not in result.stdout
    assert not env_path.exists()  # only the EXIT trap could have done this
```

Verified against real bash: exits 1, `unreachable` is not printed, file is gone.

### 3. Env var name validation

The PR description lists "Validate environment variable names" as a deliverable, but `_ENV_VAR_NAME_RE` can be deleted with a green suite.

```python
@pytest.mark.parametrize("name", ["BAD-NAME", "1FOO", "A B", "", "FOO=x", "FOO;id"])
def test_rejects_invalid_env_var_names(self, name):
    with pytest.raises(ValueError):
        _serialize_env_vars({name: "value"})
```

### 4. The in-chain `rm -f` — arguably an equivalent mutant

Removing it doesn't change the end state, because the trap still cleans up at shell exit. It changes *when*: the file would survive for the entire lifetime of the automation instead of milliseconds. If that window is the point of having both, assert on the timing rather than the end state:

```python
def test_env_file_is_gone_before_the_entrypoint_runs(self, tmp_path):
    env_path = tmp_path / "private.env"
    env_path.write_bytes(_serialize_env_vars({"PRIVATE_VALUE": "secret"}))

    result = subprocess.run(
        ["bash", "-c", f"{_env_command_prefix(str(env_path))}test ! -e {env_path} && echo GONE"],
        capture_output=True,
        text=True,
        check=True,
    )

    assert "GONE" in result.stdout
```

---

### Two smaller gaps

- **No-env case is unasserted.** `env_vars=None` and `env_vars={}` should upload once and produce a command with no prefix — that's the path taken by every automation without secrets, and nothing covers it.
- **Value edge cases.** `test_env_file_is_loaded_and_removed` covers `it's $private`, which is the important one. An empty value and a multi-line value are both reachable through the file API and are free to add as `parametrize` cases on the same test.

### Not a test gap, but related

Driving `_env_command_prefix` through a real `bash` subprocess rather than asserting on the string is the right call — it's what makes #2 and #4 above two-line additions instead of a new harness. Worth extending the same idea one step further before merge: the new `chmod` / `.` / `rm` steps sit on the critical path of every automation that has env vars, and whether `chmod 600` can fail depends on whether the file-upload API and the bash session run as the same UID — something no mock can answer. One run through `scripts/test_automation.py` against a real sandbox would close that.

This comment was generated by an AI assistant on behalf of the user.

malhotra5 (approved):
LGTM 

clever design 

some mutation testing catches seem legit, would be good to examine them


also for sanity, would be nice to test this in a feature deploy before merging

Issue #251, opened 2026-07-21
Add retention/purge for terminal automation_runs


aivong-openhands:
## Summary
`automation_runs` grows without bound — terminal runs (`COMPLETED`/`FAILED`) are never purged. The table is currently ~176k rows, of which only a handful are active (`RUNNING`); the rest are historical terminal runs.

## Why it matters
Every query over `automation_runs` gets more expensive as the table grows — the dispatcher/watchdog polls and the list-runs queries alike (see #157, #250). Unbounded growth also means slower autovacuum, more dead tuples, and larger backups. The `automation_id` index in #250 fixes the immediate sequential-scan cost, but retention is what keeps all access patterns cheap over time instead of steadily degrading.

## Proposed fix
Introduce a retention policy for terminal runs:
- Periodically delete (or archive) `COMPLETED`/`FAILED` runs older than a configurable window.
- Always keep active runs (`PENDING`/`RUNNING`) and recent terminal runs.
- Implement as a scheduled cleanup task (the service already runs a scheduler) or a periodic batched `DELETE`, sized to avoid long locks on the live table.

## Open questions
- Retention window — 30 / 60 / 90 days?
- Hard delete vs archive — do users need long-term run history through the API/UI?
- Whether partitioning by time is worth it if plain retention proves insufficient.

Refs #157, #250.

PR #277, opened 2026-07-30
fix(automation): purge expired local-mode run workspaces

### LINKED ISSUE #271: Purge terminal local-mode automation workspaces after configurable retention period
## Problem

When automation runs use the persistent local agent-server backend, each run gets a workspace under:

```text
${AUTOMATION_WORKSPACE_BASE}/automation-runs/<run_id>
```

After a run reaches a terminal state, its workspace is retained indefinitely. `LocalAgentServerBackend.cleanup_after_verification()` currently does not remove it, and there is no scheduled/startup reaper for old run workspaces. Over time this leaves hundreds of completed/failed workspaces and consumes multiple GiB of disk space.

This is filesystem retention, separate from retention of rows in the `automation_runs` database table discussed in #251.

## Expected behavior

Local-mode automation workspaces should have a configurable retention policy and be purged safely once their runs are terminal.

## Suggested behavior / acceptance criteria

- Never remove workspaces for pending or running runs.
- Remove workspaces for completed, failed, or otherwise terminal runs after a configurable age.
- Provide a documented retention setting; a finite default such as 7 days would be reasonable for local development.
- Run cleanup periodically and/or during service startup so stale workspaces are eventually collected.
- Delete in bounded batches and log what was removed and how much space was reclaimed.
- Tolerate already-missing directories and partially-created workspaces.
- Keep filesystem cleanup independent from database-row retention, while allowing both policies to be configured consistently.
- Add tests covering active-run protection, retention cutoffs, and failed deletion/retry behavior.

## Related issues

- #251 covers retention/purge of terminal `automation_runs` database rows, but not their filesystem workspaces.
- #61 discusses persistent agent-server mode and explicitly leaves cleanup of working directories/packages/temp files as an open question.

trungminhdo4-glitch:
HUMAN: I reviewed the maintainer requirements and the final implementation report for this Round-4 update before publication. The code changes and local verification on this head were performed by an agent run under my direction; I did not personally execute the tests. An independent agent diff reread was completed. The final implementation consolidates workspace cleanup into the existing watchdog, keeps filesystem-first discovery and database classification, shares the local automation-runs root with the backend, preserves active-run and path-safety protections, and leaves workspace retention as the only cleanup-specific setting.

regards

## Summary

- purge expired local-mode run workspaces from the existing watchdog janitor cycle
- discover direct UUID workspaces from the filesystem first, then classify only a bounded candidate window with one database query
- rotate candidates deferred by active status, retention, or a prior delete failure behind fresh candidates
- share the complete `automation-runs` root between the local backend and watchdog

## Why

Persistent local agent-server runs leave their workspaces under
`${AUTOMATION_WORKSPACE_BASE}/automation-runs/<run_id>` indefinitely. This can
consume significant local disk space over time.

## Safety properties

- PENDING and RUNNING workspaces are never removed
- only direct, canonical UUID directories under the configured root are candidates
- root and workspace links/junctions are refused; nested links are not followed or counted
- expiry uses `(completed_at or scandir-captured root mtime) < cutoff`
- classification is limited to three times the 50-workspace deletion budget per cycle
- candidates deferred by active status, retention, or failed/refused deletion are deprioritized on the next cycle
- database sessions close before filesystem deletion, which runs off the event loop
- shutdown is checked before each deletion and during the watchdog wait
- `workspace_retention_seconds=0` disables workspace cleanup

## Tests

- `tests/test_cleaner.py`: 41 passed, 10 skipped because link/junction creation is unavailable on this Windows host
- watchdog lifecycle regressions: 2 passed
- backend/config/local-mode/storage group: 152 passed, 1 deselected; the unfiltered run also reproduced one Windows Home-resolution environment failure in `tests/test_storage.py`, outside the changed files
- Ruff check, Ruff format check, Pycodestyle, targeted Pyright, Linux-platform Pyright, compileall, and Git diff checks passed
- the repository-authoritative pre-commit command was attempted but did not reach hooks because uv/maturin could not build `litellm==1.93.0` without `gcc.exe` and `dlltool.exe`; it is not claimed green
- the current GitHub fork workflows are awaiting approval

Fixes #271

Disclosure: This contribution was prepared with AI assistance; implementation and local verification were performed by an agent run under the owner's direction.


VascoSch92 (changes requested):
hey @trungminhdo4-glitch 

Thanks for the PR, and for being upfront about the AI assistance.

First off, the deletion path itself is solid. Building the path from the UUID rather than anything user-supplied, refusing symlinks and junctions on both the root and the entry, the containment re-check after resolve, rmtree in a thread. That's the part I'd have worried about and it's fine.

My problem is with what we iterate over. Right now we walk DB rows and turn them into paths. I think it needs to be the other way around: walk `automation-runs/`, and use the DB only to decide whether each directory is safe to delete. That one change fixes three separate things.

**Orphans.** A directory whose row is gone is invisible to us, permanently. That's not theoretical: #251 is about deleting exactly the rows we key on here, so once it lands it will strand every workspace it cleans up. Same thing happens locally any time someone resets automations.db or repoints AUTOMATION_DB_URL. The issue specifically asks for the filesystem policy to be independent of the row policy, and right now it's downstream of it.

**The scan never shrinks.** Missing dirs don't consume batch budget, which is the right call on its own, but it means once we've caught up we re-walk every terminal row ever recorded, every hour, with a to_thread hop each.

**Errors starve everything behind them.** errors and refused do count against the batch and we always start from the oldest, so a handful of undeletable dirs blocks everything newer indefinitely. With batch_size=1 and one dir whose rmtree always raises PermissionError, I ran ten cycles and the newer workspace never got touched. You already solved this exact problem for `missing`.

Roughly what I have in mind:

```python
candidates = await asyncio.to_thread(_scan_candidates, runs_root, batch_size)  # {run_id: mtime}
async with session_factory() as session:
    rows = (await session.execute(
        select(AutomationRun.id, AutomationRun.status, AutomationRun.completed_at)
        .where(AutomationRun.id.in_(candidates))
    )).all()

for run_id, mtime in candidates.items():
    row = by_id.get(run_id)
    if row is None:                       # orphan
        if mtime < cutoff: doomed.append(run_id)
    elif row.status not in TERMINAL_STATES:
        continue                          # pending/running, never touched
    elif (row.completed_at or mtime) < cutoff:
        doomed.append(run_id)
```

`_scan_candidates` is just an os.scandir that skips non-directories, links and anything that doesn't parse as a UUID. `_delete_workspace` stays exactly as you wrote it. Side benefit: the session is closed before any rmtree runs, whereas today we hold one, and on SQLite a read transaction, across the whole batch.

Can you tag me once you idd the changes? 
Thanks

trungminhdo4-glitch:
@VascoSch92 Thanks for the detailed review und danke für die Zeit und konstruktive kritik. I’ve pushed the requested changes. Cleanup now discovers candidates from automation-runs/ first and uses the DB only for batched classification. Orphan workspaces are handled by mtime, active runs remain protected, the DB session is closed before deletion, and failed/refused deletions no longer consume the successful deletion budget.

I also synchronized the branch with current main. The focused cleanup tests and adjacent lifecycle/config tests pass, along with Ruff checks. PTAL when you have a chance.

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+    )
+    result.candidates_found = len(candidates)
+
+    if candidates:
+        async with session_factory() as session:
+            stmt = select(
+                AutomationRun.id,
+                AutomationRun.status,
+                AutomationRun.completed_at,
+            ).where(AutomationRun.id.in_(list(candidates)))
```
VascoSch92: **The classification query isn't actually batched.**

My sketch passed `batch_size` into `_scan_candidates`; the implementation is `_scan_candidates(runs_root)`, so this binds one parameter per directory on disk.

```python
ids = [uuid.uuid4() for _ in range(32767)]
await session.execute(select(AutomationRun.id).where(AutomationRun.id.in_(ids)))
# sqlite3.OperationalError: too many SQL variables      (32766 is fine)
```

asyncpg has the same 32767 ceiling. `purger_loop`'s `except Exception` at :374 swallows it and logs once an hour, so from that point the purge never completes a cycle — on exactly the backlog it exists to clean. A 5-minute automation reaches 32k directories in about four months.

Either chunk the IN, or bound the scan like the sketch did.

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+
+    try:
+        if not runs_root.is_dir():
+            return {}
+        resolved_root = runs_root.resolve(strict=True)
+        candidates: list[tuple[UUID, float]] = []
+        with os.scandir(runs_root) as entries:
+            for entry in entries:
+                candidate_path = Path(entry.path)
+                if entry.is_symlink() or _is_link_or_junction(candidate_path):
```
VascoSch92: **One bad dirent aborts the whole scan.**

`entry.is_symlink()` and `_is_link_or_junction()` sit above the `try` on the next line, so their OSError lands in the outer handler at :121 and returns `{}`.

```python
# 3 valid workspaces under automation-runs/
os.chmod(runs_root, 0o444)   # readable, not executable -> lstat() gets EACCES
_scan_candidates(runs_root)  # {} — not 3
```

Injecting ESTALE on 1 of 3 entries also gives 0, not 2. The comment inside that `try` says a candidate "can disappear or become inaccessible while the directory is being scanned" — which is precisely what these two guards bypass. If the cause is permanent (one root-owned dir, a stale NFS handle), cleanup is dead for the life of the deployment.

Just move both guards inside the `try`.

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+    """Return whether a terminal row or orphan is older than ``cutoff``.
+
+    A terminal row without ``completed_at`` falls back to the workspace mtime;
+    this is also the age source for orphaned workspaces.  Invalid filesystem
+    timestamps fail closed.
+    """
+    try:
+        if completed_at is not None:
+            return ensure_utc(completed_at) < cutoff
+        return datetime.fromtimestamp(mtime, tz=UTC) < cutoff
```
VascoSch92: **Top-level mtime isn't a liveness signal.**

This is my suggestion coming back to bite, so to be clear: the clock is wrong, not the approach.

```python
t0 = ws.stat().st_mtime
(ws / "repo" / "src" / "f.py").write_text("live edit")
assert ws.stat().st_mtime == t0   # holds — 200 writes, still unchanged
```

An orphan skips the `TERMINAL_STATES` guard entirely (:272-278), so a run that's still executing while its row is invisible — someone reset `automations.db`, or repointed `AUTOMATION_DB_URL` — gets `rmtree`'d underneath the agent.

Use the newest mtime in the tree, or don't purge an orphan the agent-server still has a live conversation for.

openhands/automation/app.py
```diff
@@ -163,6 +165,24 @@ async def lifespan(app: FastAPI):
     app.state.watchdog_task = watchdog_task
     logger.info("Background watchdog started")
 
+    # Purger: removes old workspace directories in local mode only
+    purger_task: asyncio.Task | None = None
+    if settings.is_local_mode:
+        purger_task = asyncio.create_task(
+            purger_loop(
+                app.state.session_factory,
+                workspace_base=os.path.expanduser(
```
VascoSch92: **Second fallback, and it disagrees with the backend's.**

`settings.workspace_base or "/workspace"` here vs `self.workspace_base or DEFAULT_LOCAL_WORKSPACE_BASE` in `backends/local.py:201`. With `AUTOMATION_WORKSPACE_BASE=` (empty — how people unset it in a .env; `workspace_base: str` has no `min_length`) the purger scans `/workspace/automation-runs` while runs land in `~/.openhands/workspaces/automation-runs`. It then logs "Purge complete: 0 deleted … (0 candidates scanned)" forever while the disk fills.

Ask the backend for the path rather than recomputing it. `os.path.expanduser` is redundant here too — `_workspace_root` already expands.

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+            status, completed_at = row
+            if status not in TERMINAL_STATES:
+                continue
+        else:
+            completed_at = None
+
+        if not _is_expired(completed_at, mtime, cutoff):
+            continue
+
+        delete_result = await asyncio.to_thread(
```
VascoSch92: **The starvation fix removed the only bound on attempts.**

```python
os.chmod(runs_root, 0o555)   # every rmtree fails; 200 expired dirs
r = await purge_terminal_workspaces(sf, base, 604800, batch_size=50)
# r.deleted == 0, r.errors == 200  -> 200 attempts + 200 WARNINGs, every cycle
```

Cheap in wall time (~0.07s), but it's an hourly 200-line log flood that grows with the backlog.

Bound total attempts separately from successes (`attempts >= batch_size * K`) — that keeps the anti-starvation property you added.

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+                AutomationRun.completed_at,
+            ).where(AutomationRun.id.in_(list(candidates)))
+            rows = (await session.execute(stmt)).all()
+        by_id = {
+            run_id: (status, completed_at) for run_id, status, completed_at in rows
+        }
+    else:
+        by_id = {}
+
+    for run_id, mtime in candidates.items():
```
VascoSch92: **The cycle can't see `shutdown_event`.**

`purge_terminal_workspaces` doesn't take it, so the checks at :352/:377 only fire between cycles. Against the real shutdown path in `app.py:223-241`, at 0.2s per `rmtree` (a git checkout plus a .venv isn't instant): the 5s `wait_for` times out, the task is cancelled, and one more workspace finishes deleting *after* cancellation — `asyncio.to_thread` isn't interruptible — while `engine.dispose()` runs.

Thread the event through and break between deletions.

openhands/automation/config.py
```diff
@@ -532,6 +539,11 @@ class ServiceSettings(BaseSettings):
     dispatcher_interval_seconds: int = 10
     dispatcher_batch_size: int = 10
     watchdog_interval_seconds: int = 60
+    purger_interval_seconds: int = Field(default=3600, gt=0)  # 1 hour
+    purger_batch_size: int = Field(default=50, gt=0)
+
+    # Workspace retention for local mode
+    workspace_retention_seconds: int = Field(default=604800, ge=0)  # 7 days
```
VascoSch92, line 591: **`ge=0` makes 0 destructive.**

`AUTOMATION_WORKSPACE_RETENTION_SECONDS=0` is accepted and deletes workspaces that finished seconds ago. `AUTOMATION_PURGER_INTERVAL_SECONDS=0` is rejected (`gt=0`) four lines up.

Someone setting 0 to switch the feature off gets the opposite of what they wanted, and there's no opt-out flag at all — this turns on for every existing local deployment on upgrade.

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+
+    candidates_found: int = 0
+    deleted: int = 0
+    missing: int = 0
+    refused: int = 0
+    errors: int = 0
+    bytes_freed: int = 0
+
+
+class DeleteOutcome(Enum):
```
VascoSch92: this should be StrEnum

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+    {
+        AutomationRunStatus.COMPLETED,
+        AutomationRunStatus.FAILED,
+        AutomationRunStatus.CANCELLED,
+        AutomationRunStatus.SKIPPED,
+    }
+)
+
+
+@dataclass
```
VascoSch92: use frozen and slots here. 

Moreover, why there are already default values here?

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
+
+from openhands.automation.models import AutomationRun, AutomationRunStatus
+from openhands.automation.utils import ensure_utc, utcnow
+
+
+logger = logging.getLogger("automation.workspace_cleaner")
+
+TERMINAL_STATES = frozenset(
```
VascoSch92: do you need this set? You can just use the class AutomationRunStatus

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+class DeleteOutcome(Enum):
+    """Outcome of one bounded workspace deletion attempt."""
+
+    DELETED = "deleted"
+    MISSING = "missing"
+    REFUSED = "refused"
+    ERROR = "error"
+
+
+@dataclass(frozen=True)
```
VascoSch92: slots

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+class WorkspaceDeleteResult:
+    outcome: DeleteOutcome
+    bytes_freed: int = 0
```
VascoSch92: why already populated?

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+                    if str(run_id) != entry.name:
+                        continue
+                    resolved_path = candidate_path.resolve(strict=True)
+                    if (
+                        resolved_path.parent != resolved_root
+                        or not resolved_path.is_dir()
+                    ):
+                        continue
+                    mtime = entry.stat(follow_symlinks=False).st_mtime
+                except (FileNotFoundError, OSError, ValueError):
```
VascoSch92: use contextlib.suppress

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+
+def _dir_size(path: Path) -> int:
+    total = 0
+    try:
+        for dirpath, _dirnames, filenames in os.walk(path):
+            for filename in filenames:
+                filepath = Path(dirpath) / filename
+                try:
+                    total += os.path.getsize(filepath)
+                except OSError:
```
VascoSch92: use contextlib suppress here

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+                    pass
+    except OSError:
```
VascoSch92: as the above comment

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+        return {}
+
+    candidates.sort(key=lambda item: (item[1], str(item[0])))
+    return dict(candidates)
+
+
+def _dir_size(path: Path) -> int:
+    total = 0
+    try:
+        for dirpath, _dirnames, filenames in os.walk(path):
```
VascoSch92: can we not use pathlib here?

VascoSch92: you never use _dirnames here

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+        if delete_result.outcome is DeleteOutcome.MISSING:
+            result.missing += 1
+            continue
+        if delete_result.outcome is DeleteOutcome.REFUSED:
+            result.refused += 1
+            continue
+        if delete_result.outcome is DeleteOutcome.ERROR:
+            result.errors += 1
+            continue
+
```
VascoSch92: use a match statement here

openhands/automation/workspace_cleaner.py
```diff
@@ -0,0 +1,385 @@
...
+                )
+        except Exception:
+            logger.exception("Error in workspace purge scan")
+
+        if shutdown_event is not None:
+            try:
+                await asyncio.wait_for(shutdown_event.wait(), timeout=interval_seconds)
+                logger.info("Workspace purger received shutdown signal, exiting")
+                break
+            except TimeoutError:
```
VascoSch92: contextlib suppress

tests/test_workspace_purger.py
```diff
@@ -0,0 +1,895 @@
+"""Tests for workspace purging of local-mode terminal runs."""
```
VascoSch92: in general you can use @pytest.mark.parametrize to write the tests

openhands/automation/workspace_cleaner.py
VascoSch92, line 1: I don't knwo if this is a good name for the script. Do you think there are better names?

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Fixed by classifying candidate IDs in chunks of 500. The regression instruments 32,767 candidates, verifies multiple queries, and asserts every query stays within the bound.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Moved every per-entry link, type, UUID, resolve, and stat operation into the isolated block. The regression makes one of three dirents raise early and verifies the other two are still returned.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Orphan liveness now uses the newest mtime across a non-following tree traversal in a worker thread. Recent nested writes retain an old-root workspace, and traversal errors, links, or invalid mtimes fail closed.

openhands/automation/app.py
trungminhdo4-glitch: Added the Local backend-owned resolve_local_workspace_base() helper and reused it in both get_work_dir() and the purger. The test covers None, empty, explicit, and tilde values.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Kept batch_size as the successful deletion budget and added a separate total attempt cap of batch_size * 3. Regressions cover the hard cap and verify an older permanent failure does not block a newer candidate.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Threaded shutdown_event into the active purge and check it before each deletion, including after worker-thread traversal returns. The regression sets shutdown after the first attempt and verifies no second deletion starts.

openhands/automation/config.py
trungminhdo4-glitch: Retention 0 now disables startup scheduling and is also a defensive no-op in direct purge and loop calls. Added config, lifecycle-gate, direct-purge, and loop regressions.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Changed DeleteOutcome to StrEnum as requested.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: PurgeResult is now frozen and slotted with required fields only. Counters accumulate locally and the complete result is constructed once at the end.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Removed TERMINAL_STATES and now match AutomationRunStatus explicitly for COMPLETED, FAILED, CANCELLED, and SKIPPED. Existing active-state and all-terminal-state tests remain green.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: WorkspaceDeleteResult is now frozen and slotted.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Removed the bytes_freed default; every deletion outcome now supplies it explicitly.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Replaced the per-entry exception-only block with contextlib.suppress while keeping the entire dirent operation isolated.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Replaced the per-file size error handler with contextlib.suppress.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Replaced the outer size-walk error handler with contextlib.suppress as well.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: _dir_size now uses pathlib paths and non-following Path.stat(), while retaining the link/reparse safety checks.

trungminhdo4-glitch: Removed the unused directory-name binding by using _ in the walk tuple.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Deletion outcome handling now uses an exhaustive match statement.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: The periodic wait now uses contextlib.suppress(TimeoutError) inside the loop.

tests/test_workspace_purger.py
trungminhdo4-glitch: Consolidated the new startup-mode cases with pytest.mark.parametrize and retained separately named safety regressions where parametrization would hide intent.

openhands/automation/workspace_cleaner.py
trungminhdo4-glitch: Renamed the module to workspace_purger.py to match its destructive operation and purger_loop terminology; imports, logger, and the test module were updated atomically.

trungminhdo4-glitch:
@VascoSch92 Thanks for the thorough follow-up. I addressed the open review threads in `02fd5cc`, including bounded DB classification, per-entry scan isolation, orphan liveness protection, the backend-owned workspace path, bounded deletion attempts, in-cycle shutdown handling, retention=0 disable semantics, and the requested cleanup/refactors. The affected suite is 155 passed / 3 Windows symlink-privilege skips; repository-wide Ruff/format, affected Pyright, compileall, and diff checks pass. A broader local run reached 942 passed before tests requiring unavailable Docker/PostgreSQL fixtures and pre-existing Windows AF_UNIX assumptions failed. The fork workflows currently have the PR check passing, with test/CI/Docker/artifact workflows awaiting maintainer approval (`action_required`). PTAL when you have a chance.

openhands/automation/workspace_purger.py
```diff
@@ -0,0 +1,403 @@
...
+        return {}
+    except OSError as exc:
+        logger.warning("Failed to scan workspace root %s: %s", runs_root, exc)
+        return {}
+
+    candidates.sort(key=lambda item: (item[1], str(item[0])))
+    return dict(candidates)
+
+
+def _latest_tree_mtime(path: Path) -> float | None:
```
VascoSch92: This is the blocking one. _latest_tree_mtime returns None the moment it sees a symlink anywhere below the workspace, and the caller then skips the directory silently. These are agent working directories — they contain .venv/bin/python, node_modules/.bin/*, git symlinks, as a matter of course.

Two identical 60-day-old orphans, one with a single venv link:

```
a = _workspace_path(base, uuid4()); (a / "repo").mkdir(parents=True)
(a / "repo" / "main.py").write_text("x" * 100)

b = _workspace_path(base, uuid4()); (b / "repo" / ".venv" / "bin").mkdir(parents=True)
(b / "repo" / "main.py").write_text("x" * 100)
os.symlink(sys.executable, b / "repo" / ".venv" / "bin" / "python")

# both aged to 60 days, retention 7 days:
# PurgeResult(candidates_found=2, deleted=1, ...)
# a is gone. b is still on disk, and nothing was logged about it.
```

So orphan collection works on empty directories and not on the ones actually holding the multiple GiB in #271.

The fail-closed reasoning doesn't hold either: rmtree doesn't follow symlinks, and you already re-check containment on the workspace root before deleting, so a link inside the tree can't cause an escape. Refusing to follow it is right; refusing to compute an mtime because one exists is not.

My preference is to delete this and _valid_mtime outright rather than patch them. That's 38 lines plus a thread hop and a full recursive stat per orphan candidate, every cycle, to answer a question the scandir mtime already answers well enough. The only candidates that reach this branch have no DB row at all, or a terminal row with no completed_at — an orphan by definition has no live run, and the root mtime against a 7-day window is the guard. (completed_at or mtime) < cutoff is the whole rule.

openhands/automation/workspace_purger.py
```diff
@@ -0,0 +1,403 @@
...
+from openhands.automation.models import AutomationRun, AutomationRunStatus
+from openhands.automation.utils import ensure_utc, utcnow
+
+
+logger = logging.getLogger("automation.workspace_purger")
+
+# 500 stays below SQLite's historical 999-variable default as well as asyncpg's
+# much larger parameter ceiling, leaving room for future query predicates.
+DB_CLASSIFICATION_CHUNK_SIZE = 500
+DELETE_ATTEMPT_FACTOR = 3
```
VascoSch92: use Final from typing

openhands/automation/cleaner.py
VascoSch92, line 1: change the name of the script. Like `cleaner.py`

openhands/automation/workspace_purger.py
```diff
@@ -0,0 +1,403 @@
...
+        age: datetime | float
+        if completed_at is not None:
+            age = completed_at
+        else:
+            if not _is_expired(root_mtime, cutoff):
+                continue
+            tree_mtime = await asyncio.to_thread(
+                _latest_tree_mtime, _workspace_path(workspace_base, run_id)
+            )
+            if tree_mtime is None:
```
VascoSch92: Bare continue with no log, so a workspace skipped here is indistinguishable from one that was purged — the PurgeResult counters don't record it either. See my comment on _latest_tree_mtime; if that helper goes away this does too, but if you keep any version of it, an unpurgeable candidate needs to be visible somewhere.

openhands/automation/workspace_purger.py
```diff
@@ -0,0 +1,403 @@
...
+            case DeleteOutcome.MISSING:
+                missing += 1
+            case DeleteOutcome.REFUSED:
+                refused += 1
+            case DeleteOutcome.ERROR:
+                errors += 1
+            case DeleteOutcome.DELETED:
+                deleted += 1
+                bytes_freed += delete_result.bytes_freed
+                logger.debug(
```
VascoSch92: This is the line that says what was actually removed, and it's at debug — while the summary that usually reads 0 deleted, 0 missing, 0 refused, 0 errors is at info. #271 explicitly asks for what was removed and how much space was reclaimed to be logged. The two levels are the wrong way round.

openhands/automation/workspace_purger.py
```diff
@@ -0,0 +1,403 @@
...
+                    "Purged workspace for run %s (%d bytes freed)",
+                    run_id,
+                    delete_result.bytes_freed,
+                )
+
+    result = PurgeResult(
+        len(candidates), deleted, missing, refused, errors, bytes_freed
+    )
+    logger.info(
```
VascoSch92: Fires unconditionally, so every idle local deployment gets Purge complete: 0 deleted, 0 missing, 0 refused, 0 errors, 0 bytes freed (0 candidates scanned) once an hour, forever. watchdog_loop guards its equivalent with if pruned: for exactly this reason — please match it and drop the no-op case to debug.

openhands/automation/workspace_purger.py
```diff
@@ -0,0 +1,403 @@
...
+    return True
+
+
+def _dir_size(path: Path) -> int:
+    total = 0
+    with suppress(OSError):
+        for dirpath, _, filenames in os.walk(path, followlinks=False):
+            for filename in filenames:
+                filepath = Path(dirpath) / filename
+                if _is_link_or_junction(filepath):
```
VascoSch92: Redundant: os.walk(..., followlinks=False) already won't descend symlinked directories, and stat(follow_symlinks=False) already returns the link's own size rather than the target's. So this costs two extra stat syscalls per file to avoid counting a handful of bytes.

Worth a wider look while you're here: _dir_size is a full extra tree walk that rmtree immediately repeats, purely to produce a number for a log line. On a multi-GiB workspace that's real I/O. I'm fine keeping it — #271 asks for reclaimed space — but let's not pay three stats per file for it.

openhands/automation/workspace_purger.py
```diff
@@ -0,0 +1,403 @@
...
+    return total
+
+
+def _is_link_or_junction(path: Path) -> bool:
+    """Return whether path is a symlink or Windows directory junction."""
+    is_junction = getattr(path, "is_junction", None)
+    if path.is_symlink() or (is_junction is not None and is_junction()):
+        return True
+    try:
+        attributes = getattr(path.stat(follow_symlinks=False), "st_file_attributes", 0)
```
VascoSch92: requires-python = ">=3.12" and Path.is_junction() landed in 3.12, so getattr(path, "is_junction", None) never returns None and this branch is effectively unreachable for the case it's written for. The three tests covering it (test_skips_detected_reparse_point, test_refuses_detected_junction, test_refuses_detected_workspace_root_junction) monkeypatch the thing under test, so they're asserting the mock. I'd drop the fallback and the tests.

openhands/automation/workspace_purger.py
```diff
@@ -0,0 +1,403 @@
...
+    session_factory: async_sessionmaker[AsyncSession],
+    workspace_base: str | os.PathLike[str] | None,
+    retention_seconds: int,
+    batch_size: int = 50,
+    shutdown_event: asyncio.Event | None = None,
+) -> PurgeResult:
+    """Purge expired terminal and inactive orphan workspace directories.
+
+    ``batch_size`` limits successful deletions. Total attempts are separately
+    limited to three times that value so failures cannot create an unbounded
```
VascoSch92: The docstring overstates what this buys. It bounds the warning storm per cycle, but it doesn't remove starvation: with more than 3 × batch_size permanently-undeletable directories, every attempt goes to failures — zero deletions and 150 warnings per cycle, indefinitely.

openhands/automation/workspace_purger.py
```diff
@@ -0,0 +1,403 @@
...
+    match status:
+        case (
+            AutomationRunStatus.COMPLETED
+            | AutomationRunStatus.FAILED
+            | AutomationRunStatus.CANCELLED
+            | AutomationRunStatus.SKIPPED
+        ):
+            return True
+        case _:
+            return False
```
VascoSch92: this is not a good use of `match` it is just an `in` assertion.

Please re-read your code after is generated.

tests/test_workspace_purger.py
```diff
@@ -0,0 +1,1082 @@
...
+            workspace_base="/workspace",
+            retention_seconds=0,
+            interval_seconds=3600,
+        )
+
+        mock_purge.assert_not_awaited()
+
+
+class TestPurgeResult:
+    def test_partial_success(self):
```
VascoSch92: this is testing python not the code here.

tests/test_workspace_purge.py
VascoSch92, line 1: You can really delete a lot from this test script.

VascoSch92 (commented):
Please re-read the code after it is AI-generated.

openhands/automation/workspace_purger.py
trungminhdo4-glitch: Reproduced before the fix: two expired orphan workspaces aged past retention, only the link-free one collected (deleted=1 of candidates_found=2), with nothing logged or counted for the other.

The fix deletes `_latest_tree_mtime` and `_valid_mtime` outright. Expiry is now `(completed_at or scandir-captured root mtime) < cutoff`, with the root mtime captured once at scan time, so internal symlinks/junctions no longer make expired orphans immortal. Deletion still does not follow nested link targets, still re-checks containment against the resolved root, and still refuses linked roots and paths.

Wording we can stand behind: normal service-owned live runs retain their AutomationRun row and no separate authoritative live-workspace registry exists, so the retention-window root mtime is the fallback guard for DB-less candidates, as you proposed. Regressions cover a normal internal link, a broken link, and a real directory junction.

openhands/automation/workspace_purger.py
trungminhdo4-glitch: Both constants are annotated `Final[int]` now, consistent with the existing style in git_sync.

openhands/automation/cleaner.py
trungminhdo4-glitch: Renamed to `cleaner.py` via git mv; the test module and imports follow, and the logger is now `automation.cleaner`. No alias kept — the application import was the sole consumer.

openhands/automation/workspace_purger.py
trungminhdo4-glitch: Gone with the removed recursive-mtime branch; the surviving skips are expected states already reflected in candidates_found.

openhands/automation/workspace_purger.py
trungminhdo4-glitch: Levels swapped: successful deletion logs at INFO with run id and bytes freed, and the aggregate summary is INFO only when something was deleted.

openhands/automation/workspace_purger.py
trungminhdo4-glitch: Idle cycles drop to DEBUG now, matching watchdog_loop's guarded pattern. Failures stay visible through the per-item warnings in _delete_workspace.

openhands/automation/workspace_purger.py
trungminhdo4-glitch: Redundant per-file link checks are gone; one non-following stat per file remains so external targets are not counted as reclaimed bytes.

Pre-push follow-up on your wider point: I measured _dir_size against a real junction and os.walk does descend Windows junctions even with followlinks=False — a workspace holding one junction to a 100k target reported 100010 bytes instead of 10. Linked directories are now pruned from dirnames before descent (one check per directory, none reintroduced per file), which brings it back to 10.

openhands/automation/workspace_purger.py
trungminhdo4-glitch: Fallback removed; _is_link_or_junction is now path.is_symlink() or path.is_junction(). The three monkeypatch tests are deleted and replaced with real-junction behavior tests created via _winapi.CreateJunction (junctions need no admin rights); on platforms without junction support they skip rather than assert the mock.

openhands/automation/workspace_purger.py
trungminhdo4-glitch: Docstring rewritten: it bounds attempts and log volume per cycle only and explicitly disclaims any global anti-starvation guarantee.

openhands/automation/workspace_purger.py
trungminhdo4-glitch: Replaced with a plain membership expression over AutomationRunStatus.COMPLETED / FAILED / CANCELLED / SKIPPED.

tests/test_workspace_purger.py
trungminhdo4-glitch: Deleted — it only asserted that a frozen dataclass constructor returns the fields it was given.

tests/test_workspace_purge.py
trungminhdo4-glitch: Done: 49 to 52 collected with less mock surface. Removed seven Python/mock/detail tests (dataclass round-trip, three junction-mock tests pinning the deleted fallback, two tree-liveness mock tests, and an orphan mtime test your suggested design intentionally inverts), added real-behavior regressions for internal links, external-target preservation, real junctions, deletion-evidence logging at INFO, and idle-cycle DEBUG.

trungminhdo4-glitch:
@VascoSch92 Thanks for the follow-up. I re-read the full cleanup path and addressed the latest review round on the new head.

The blocking internal-link case was reproduced before the fix and is now covered by regression tests. The recursive tree-mtime helpers are gone; orphan fallback uses the scandir-captured root mtime, while the existing root containment/link/junction deletion protections remain intact.

I also applied the requested simplifications around Final, module naming, logging, Path.is_junction(), terminal-state membership, attempt-cap wording, and test cleanup. During the final pre-push review I found and fixed one additional Windows junction byte-accounting edge case (junction targets were counted as reclaimed bytes; they are now pruned before descent).

Focused and adjacent validation is green; the PR description contains the current test evidence and environment limitations.

PTAL when you have a chance.

VascoSch92 (changes requested):
hey @trungminhdo4-glitch

Thanks for turning this around so fast, and for the filesystem-first rewrite. That's exactly what I was after: scandir first, DB only to classify, session closed before any rmtree, orphans handled by mtime. The deletion path is good and I don't want you to touch it — the UUID-derived path, the link refusal on both root and entry, the containment re-check after resolve, and `completed_at or mtime` for expiry all stay as they are.

Before you do another round though, I owe you a scope call I should have made earlier, because it's structural and it would be unfair to have you polish inside a shape I'm going to ask you to change.

**Fold this into the watchdog instead of adding a fourth loop.** `watchdog.py` says it in its own docstring: *"The same loop prunes `integration_events`: it is the service's only periodic janitor, and a second loop buys nothing for one bounded DELETE."* The workspace purge is the same kind of work, so I'd like `purge_terminal_workspaces` to live in `watchdog.py` as a third janitor step next to `prune_integration_events`, with `cleaner.py`, `purger_loop` and the lifespan task all going away. The first watchdog cycle runs before the first sleep, so you keep the startup purge for free, and the shutdown/`await` handling in `app.py` reverts to what it was.

That also collapses the config surface from three knobs to one. `purger_interval_seconds` disappears with the loop, and `purger_batch_size` becomes a module constant next to `PRUNE_BATCH_SIZE`, which is how the existing pruner is bounded:

```python
WORKSPACE_PURGE_BATCH_SIZE = 50
```

leaving just `workspace_retention_seconds`. Minor, non-blocking: I'd take days over seconds so it reads alongside `integration_event_retention_days`, but if you feel strongly about seconds being easier to test, keep them.

If you'd rather argue for a separate loop — the 1h vs 60s cadence is a fair point — then it has to use `run_periodic_loop` from `utils/periodic_loop.py`. It already does shutdown-event exit, per-cycle exception isolation and the `wait_for`/`TimeoutError` sleep, and `purger_loop` currently reimplements all three. Its docstring literally flags this as the gap it exists to close.

The rest is small:

**`automation-runs` is a literal in two modules now.** `resolve_local_workspace_base` only factors out the expanduser, so `cleaner.py::_workspace_root` and `local.py::get_work_dir` each append the segment themselves and can drift. Push the helper one level up so it owns the whole runs root and both callers join onto it:

```python
def local_runs_root(workspace_base: str | os.PathLike[str] | None) -> Path:
    return Path(resolve_local_workspace_base(workspace_base)) / "automation-runs"
```

Keep the `resolve(strict=False)` on the cleaner side rather than baking it into the helper — `get_work_dir` should keep returning the path the user configured, and the containment check wants the resolved one.

**Bound the classification by the batch.** Right now every directory found gets classified — 10k workspaces is 20 chunked queries per cycle to delete 50. Sort the candidates, slice to the attempt cap, then run one `IN` query. `batched` and `DB_CLASSIFICATION_CHUNK_SIZE` both go away.

**Starvation.** The 3x attempt factor bounds the cost of a bad cycle, which is an improvement, but oldest-first ordering means persistently undeletable directories sit at the front of every cycle and can still consume the whole budget — your docstring says as much. Since you're sorting anyway, carry a small set of run ids that failed on the previous cycle and sort those last:

```python
candidates.sort(key=lambda item: (item[0] in failed_last_cycle, item[1], str(item[0])))
```

The watchdog can own that set. Five lines, and it closes the case I raised in round 1.

**CI is red on the junction test helper.** `tests/test_cleaner.py:46` — `"CreateJunction" is not a known attribute of module "_winapi"` under pyright, which is the only thing failing `backend`. Guarding that import fixes the build. Separately, and this is your call rather than a blocker: we ship a Linux container and `is_junction()` is always False there and on macOS, so I'd be comfortable dropping the junction handling and its ~10 skipped tests entirely. If you'd rather keep the prod-side refusal as cheap insurance, that's fine by me — just get the test helper past pyright.

I know that's another round after three already, and the structural one is on me for not calling it sooner. Everything above should be subtraction rather than new work, and I'd expect the diff to come down a fair bit — most of `cleaner.py` moves or disappears rather than getting rewritten.

Tag me when it's up and I'll turn it around quickly.

Thanks

trungminhdo4-glitch:
@VascoSch92 Round 4 is up on ce2777a. The final multi-cycle audit reproduced a bounded-window starvation case where active candidates could hide an eligible workspace; the watchdog now deprioritizes candidates deferred by the previous cycle while keeping one bounded IN query and active-run protection. The Linux Pyright junction-helper issue is fixed, and focused/static checks are green within the documented local limits. Fork workflows are awaiting approval. PTAL.

VascoSch92 (changes requested):
hey @trungminhdo4-glitch

The structural round landed well. `cleaner.py` is gone and the purge sits next to `prune_integration_events` as a third janitor step, the config surface is down to just `workspace_retention_seconds`, `local_runs_root` owns the path with the `resolve(strict=False)` kept on the watchdog side where the containment check needs it, and the classification is one bounded `IN` query with `batched` and the chunk constant deleted. That's all of it, thanks.

One blocker on the new rotation, and I'm afraid it doesn't do what the commit message says.

**The deferred set is replaced rather than accumulated, so it oscillates between two windows.**

```python
deferred_last_cycle.clear()
deferred_last_cycle.update(deferred_this_cycle)
```

The sort key only ever knows about the *previous* cycle, so cycle N+2 has already forgotten what cycle N deferred and walks straight back into it. Nothing ever advances past the second window.

I ran your exact key and slice over 1000 candidates where the oldest 300 are persistently not actionable and one eligible workspace sits at position 900:

```
cycle 1: window = [0..149]
cycle 2: window = [150..299]
cycle 3: window = [0..149]
cycle 4: window = [150..299]
...
after 500 cycles: the eligible workspace was never considered
```

Changing the one line to accumulate — `deferred_last_cycle |= deferred_this_cycle`, clearing only once a sweep completes — reaches it on cycle 7. A persistent offset cursor into the sorted list would work too, and is arguably simpler to reason about than a growing set.

To be fair about the preconditions: this needs more than two windows' worth of persistently non-actionable candidates sorting older than an eligible one, so undeletable directories or long-lived active runs rather than the everyday case, because expiry is mostly monotonic in mtime and the deletable ones normally sort to the front on their own. But it is exactly the guarantee `rotate deferred workspace candidates` claims to provide, so I'd rather it actually held.

The rest is tidying:

**`tests/test_cleaner.py` outlived `cleaner.py`.** 1212 lines of tests for code that now lives in `watchdog.py`, in a file named after a module that no longer exists. Merge it into `test_watchdog.py`, or rename it to something like `test_workspace_purge.py` if you'd rather keep it separate — I don't mind which, but the current name will mislead whoever greps for it next.

**`attempts >= max_attempts` is now dead.** `ordered_candidates` is already sliced to `max_attempts`, so the guard can never fire before the loop runs out. Related: `DELETE_ATTEMPT_FACTOR` no longer bounds attempts, it sizes the classification window — worth renaming to match what it does now.

**Leftover scaffolding in `app.py`.** The `list[tuple[str, asyncio.Task | None]]` annotation and the `if task is None: continue` were there for the purger task that no longer exists; no entry can be `None` any more. Revert that hunk to the original `background_tasks` block so the diff stays limited to the comment change.

Non-blocking: `purge_terminal_workspaces` takes `workspace_base` and `retention_seconds` while its neighbour `prune_integration_events` takes `settings`. Taking `settings` would read more consistently, but I won't hold the PR for it.

Last thing — CI hasn't actually run on this head, the fork workflows are still pending approval, so the Pyright junction fix is unverified. I'll approve the run so we get a real signal.

Fix the rotation and I think this is done.

Thanks

VascoSch92 (approved):
I fixed some nits but it seems good to me.

thanks

PR #280, opened 2026-07-30
feat: record accumulated LLM cost per automation run

### LINKED ISSUE #242: Record accumulated LLM cost per automation run
## Summary

Automation runs currently don't record the cost of the conversation they execute. The `accumulated_cost` is already available inside the sandbox via the SDK's `conversation_stats`, and the preset scripts even print it to stdout, but it's never sent back in the completion callback or persisted to the database.

Adding cost tracking per run would enable users to understand their automation spending, identify expensive automations, and make informed decisions about optimization (e.g., switching high-frequency deterministic tasks away from LLM presets).

## Current Behavior

- The preset SDK scripts (`presets/prompt/sdk_main.py:370-371`, `presets/plugin/sdk_main.py:421-422`) read `conversation.conversation_stats.get_combined_metrics().accumulated_cost` and print it, but discard the value.
- `RunCompleteRequest` (`schemas.py:614`) only carries `status`, `run_id`, `conversation_id`, and `error`.
- The `AutomationRun` model (`models.py:121`) has no cost/price column.
- The completion callback endpoint (`router.py`) does not extract or store any cost data.

### 1. Model — add a `cost` column to `AutomationRun`

Add a nullable `Float` (or `Numeric`) column to store the accumulated cost in USD:

```python
cost = Column(Float, nullable=True, doc="Accumulated LLM cost in USD for this run")
```

A new Alembic migration would add this column (cross-database compatible per existing conventions — no PostgreSQL-specific types).

### 2. Schema — extend `RunCompleteRequest` to accept `cost`

```python
class RunCompleteRequest(BaseModel):
    status: Literal["COMPLETED", "FAILED"]
    run_id: str | None = None
    conversation_id: str | None = None
    error: str | None = None
    cost: float | None = None  # new
```

### 3. Schema — expose `cost` in `AutomationRunResponse`

Add a `cost: float | None` field so the runs list and detail endpoints return it.

### 4. Callback endpoint — persist cost on completion

In the `complete_run` handler in `router.py`, save `request.cost` to the run record when provided.

### 5. Preset scripts — pass cost in the callback

Update both `presets/prompt/sdk_main.py` and `presets/plugin/sdk_main.py` to include `accumulated_cost` in the callback payload instead of only printing it.

### 6. Migration

Add an Alembic migration to add the `cost` column. Should follow the cross-database compatibility pattern documented in AGENTS.md (use generic `Float`, not PostgreSQL-specific types).

## Out of Scope (future work)

- Aggregated cost dashboards or per-automation cost summaries
- Cost limits / budget alerts per automation
- Token-level breakdown (input vs output tokens, per-model costs)
- Cost tracking for custom script automations that don't use preset SDK scripts

---

hieptl:
HUMAN:

I have verified the changes.

---

AGENT:

## Why

Automation runs don't record what they cost, so users can't see what their automations spend or spot the expensive ones. The value already exists inside the sandbox — both preset scripts compute `conversation.conversation_stats.get_combined_metrics().accumulated_cost` and print it — but it's a local variable that gets thrown away, and `automation_runs` has no column to hold it.

One correction to the issue as filed: it proposed making the preset scripts add `cost` to the callback payload, but **the preset scripts never build that payload** — the SDK does, in `Workspace.__exit__` → `_send_completion_callback()`. The transport half therefore lands in `software-agent-sdk`; this PR is the service half plus the preset-side call.

## Summary

- Add a nullable `cost` column to `AutomationRun` (generic `sa.Float`, cross-database per `AGENTS.md`) with migration `012`; accept `cost` on `RunCompleteRequest` and expose it on `AutomationRunResponse`.
- `complete_run` persists it on **both** `COMPLETED` and `FAILED` — failed runs still spend money.
- Both preset scripts read the cost in the `finally` block and hand it to the workspace via `register_cost()`, so failed runs report their spend too. The read is wrapped in `try/except`: it falls back to the agent server when no state is cached, and on the failure path that server may already be gone, so it must never mask the original exception.

## Issue Number

Resolves https://github.com/OpenHands/automation/issues/242

## How to Test

Automated:

```bash
uv run pytest tests/test_router.py -k CompleteRun -q
uv run pytest tests/test_db.py -k Migrations -q

```

Red/green was verified explicitly — with the source changes stashed and migration `012` moved aside, the new tests fail:

```
FAILED tests/test_router.py::TestCompleteRun::test_complete_run_saves_cost
FAILED tests/test_router.py::TestCompleteRun::test_complete_run_saves_cost_for_failed_runs
FAILED tests/test_router.py::TestCompleteRun::test_complete_run_without_cost_leaves_it_unset
================= 3 failed, 118 deselected, 1 warning in 2.98s =================

tests/test_db.py: assert "cost" in run_columns
E   AssertionError: assert 'cost' in {'automation_id', 'bash_command_id', 'completed_at',
                                      'conversation_id', 'created_at', 'error_detail', ...}

```

Full suite with the change applied:

```
$ uv run pytest tests/ -q
================= 1066 passed, 10 warnings in 70.76s (0:01:10) =================

```

Migration verified by hand on real SQLite, **both directions**:

```
$ AUTOMATION_DB_URL="sqlite+aiosqlite:///$D" uv run alembic upgrade head
--- columns after upgrade ---
[('cost', 'FLOAT')]
head: [('012',)]
--- downgrade -1 ---
cost present after downgrade: False
head: [('011',)]

```

End-to-end HTTP round trip against the real ASGI app (in-memory SQLite, `get_session` / `authenticate_request` overridden the same way `tests/conftest.py` wires them):

```
COMPLETED  -> response cost=0.4213  db cost=0.4213
FAILED     -> response cost=1.5
omitted    -> 200, cost stays NULL (back-compat with SDK 1.37.0)
list runs  -> costs=[0.4213, 1.5]

```

That third line is the compatibility case that matters: the currently pinned SDK (1.37.0) sends no `cost` field, and the callback must keep succeeding.

Preset scripts: `python -m py_compile` clean, both files byte-identical in the changed region (`diff` of the `finally` block), and every line under the file's de-facto 88-char limit.

Lint/format: `uv run ruff check openhands/ migrations/ tests/` and `ruff format --check` clean.

## Video/Screenshots

N/A — backend only, no UI in this PR. Evidence is the round-trip and migration transcripts above.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


openhands/automation/presets/prompt/sdk_main.py
```diff
@@ -384,10 +384,18 @@ def event_callback(event) -> None:
...
         print(f"  events received: {len(received_events)}")
     finally:
+        # Report cost even when the run failed — failed runs still spend.
+        # This read falls back to the agent server when no state is cached,
+        # and on the failure path the server may already be gone, so never
+        # let it mask the original exception.
+        try:
+            stats = conversation.conversation_stats.get_combined_metrics()
+            print(f"  cost: {stats.accumulated_cost}")
+            workspace.register_cost(stats.accumulated_cost)
```
malhotra5: I think the workspace registering cost should be baked into the sdk layer, we shouldn't rely on users baking this logic into their automation scripts

WDYT?

malhotra5 (changes requested):
Left a comment regarding the workspace API for registering cost

openhands/automation/presets/prompt/sdk_main.py
hieptl: Hello @malhotra5,

I have updated the code based on your feedback.

Thank you very much! 🙏

tofarr (approved):
🍰

DevinVinson:
Thank you @hieptl !

Do you want a chance to give this another 👀 @malhotra5 ?

malhotra5 (approved):

PR #282, opened 2026-07-30
feat: add explicit conversation links to automation activity log

### LINKED ISSUE #281: Make automation activity-log conversation links explicit
## Summary

Automation runs that create a conversation should expose a clear, discoverable link to that conversation in the activity log.

## Problem

The current activity-log entry is clickable as a whole, but it does not visibly communicate that it navigates to a conversation. This makes the destination easy to miss and is less clear for keyboard and screen-reader users.

## Proposed change

Show a visible **View conversation** link for each run with a `conversation_id`, preserving the existing URL behavior and the non-interactive state for runs that did not create a conversation.

## Acceptance criteria

- [ ] Each activity-log run with a conversation displays an explicit conversation link
- [ ] The link opens the corresponding conversation in a new tab
- [ ] Runs with no conversation remain clearly non-interactive
- [ ] Component tests cover both states
- [ ] PR includes a screenshot of the updated activity log

---
Created on behalf of the user.

neubig:
## Summary

Closes #281.

Adds a visible, localized **View conversation** link next to the status of every automation run that has a conversation. Runs without a conversation remain non-interactive and retain their existing explanatory state.

## Why

Activity-log rows previously relied on the whole row being clickable, which did not clearly communicate the conversation destination. The explicit link makes the navigation discoverable while retaining the existing deployment-aware conversation URL behavior.

## Screenshot

Automation activity log showing View conversation links

## Validation

- `npm test -- --run src/__tests__/components/automations/detail/activity-log-item.test.tsx`
- `npm run lint`

## Remediation blocker (2026-08-08 remediation pass)

This PR is **superseded**: main commit `35870e2 feat: retire the standalone automation frontend (#284)` deleted the entire `frontend/` tree this PR modifies. Merging current main leaves a single leftover `frontend/docs/screenshots/activity-log-conversation-link.png` and no functional diff. The "View conversation" link feature cannot be validated or landed here in its current form — it would need to be ported to wherever the automation UI now lives. No code changes were pushed.

## Live evidence

- Exercised the automation completion endpoint against a real ephemeral PostgreSQL container: completed and failed runs persisted their conversation IDs (`uv run pytest tests/test_router.py -k complete_run -q -rA`: 7 passed).


neubig:
The current diff contains only an orphan screenshot; the functional activity-log link implementation was removed during conflict resolution. This PR should be closed or rebuilt against the current automation UI rather than merged as-is. This status was prepared by an AI agent (OpenHands) on behalf of Graham Neubig.

PR #283, opened 2026-07-30
fix(scheduler): skip cron runs while an automation is active

### LINKED ISSUE #218: Add a minimal per-automation overlap guard (skip when a run is already in-flight)
## Context
Follow-up to #212, and a **minimal, shippable subset of #72** (hierarchical concurrency
limits). #72 is a broad/stale umbrella; this issue carves out just the single safeguard
needed to remove the sharp edge that #212's higher ceiling amplifies. Also related to the
queue-based alternatives in #19 / #74.

## Problem
Nothing prevents a single automation from overlapping itself:
- `is_automation_due` (`openhands/automation/utils/cron.py`) is purely cron-vs-clock and
  uses `last_triggered_at`, which is stamped at **schedule** time in `create_pending_run`,
  not at completion — so a still-running prior run never defers the next fire.
- `create_pending_run` (`openhands/automation/utils/run.py:79-112`) does not check for an
  existing PENDING/RUNNING run of the same automation.
- The dispatcher (`openhands/automation/dispatcher.py` `_poll_pending_runs` /
  `dispatch_pending_runs`) claims any row with `status == PENDING` and never checks for an
  already-active run of the same `automation_id`.

Result: a run that outlives its interval stacks concurrent runs of the same automation,
each holding its own sandbox. Raising `max_run_duration` from 10 -> 30 min triples the
maximum overlap for a given schedule.

## Minimal approach
Enforce an effective per-automation concurrency of 1 (the `default_per_automation = 1`
from #72) **without** the org/billing hierarchy:
- At schedule time (cron path): skip creating a new PENDING run if an active
  (PENDING or RUNNING) run already exists for the automation. Manual `/dispatch` may opt
  to bypass.
- Optionally also guard at dispatch claim time as defense-in-depth.
- Single query on `automation_runs(automation_id, status)` — pairs with the missing index
  in #157.

## Scope / non-goals
- Not the full #72 hierarchy (org aggregate caps, per-automation configurable limits,
  billing). Keep #72 open for that.
- Make the guard overridable so users who genuinely want overlap can opt out later.

---
_This issue was filed by an AI agent (OpenHands) on behalf of the team, as a follow-up to the review discussion on #212._

trungminhdo4-glitch:
HUMAN: I reviewed the scheduler changes and tested the overlap guard locally. I checked that cron runs are skipped when the same automation already has a PENDING or RUNNING run, while terminal runs still allow the next cron run. I also checked the batch rotation, telemetry and timestamp behavior, and that manual/event triggered dispatch is not affected.
Summary

- skip cron-triggered run creation when the same automation already has a
  `PENDING` or `RUNNING` run
- preserve normal scheduling after terminal runs
- keep manual and event-triggered dispatch behavior unchanged
- preserve scheduler polling fairness and telemetry semantics

## Why

A cron automation can currently schedule another run while its previous run is
still pending or running. Long-running automations can therefore stack multiple
concurrent sandboxes.

## Concurrency contract

The guard is evaluated inside the scheduler transaction that already locks the
automation row with `FOR UPDATE SKIP LOCKED` on PostgreSQL.

This PR intentionally applies the guard only to cron scheduling. Manual and
event-triggered dispatch behavior remains unchanged, and no database uniqueness
constraint is introduced, leaving room for an explicit per-automation override
later.

SQLite remains covered under the repository's existing single-process
assumption.

## Behavior

Blocking statuses:

- `PENDING`
- `RUNNING`

Non-blocking statuses:

- `COMPLETED`
- `FAILED`
- `CANCELLED`
- `SKIPPED`

When a run is skipped:

- `last_polled_at` is still updated
- `last_triggered_at` is unchanged
- no run row is created
- no scheduled/created telemetry event is emitted

## Tests

- scheduler regression tests covering active and terminal statuses
- mixed-automation, single-batch-query, and batch-rotation cases
- telemetry and timestamp semantics
- Ruff check
- Ruff format check
- Python compilation check

The functional behavior was tested with in-memory SQLite. PostgreSQL
multi-worker behavior was not exercised against a live PostgreSQL instance; the
serialization property follows the scheduler's existing automation-row lock.

Fixes #218

Disclosure: This contribution was prepared with AI assistance and independently
reviewed and tested locally before publication.


trungminhdo4-glitch:
HUMAN: (Owner-authorized; posted by my agent on my behalf.)

I am supervising this contribution, and it remains a draft by design. Before I decide it is ready for review, I want the concurrency contract to remain explicit: one batch query for due automations; `PENDING` and `RUNNING` block cron scheduling only; a skip updates `last_polled_at` without changing `last_triggered_at` or emitting scheduled/created telemetry; PostgreSQL relies on the existing automation-row lock, while SQLite retains its single-process assumption. The two lightweight PR checks pass; the four full fork workflows remain `action_required` and require maintainer approval before they can run. I will make the final ready-for-review and requested-change decisions.

PR #296, opened 2026-08-01
feat: report the configured automation timeout cap

neubig:
## Summary

- expose the configured `max_run_duration` as `maxAutomationTimeoutSeconds` in capability discovery
- return the same runtime-derived value for ready and unavailable deployments
- cover an overridden deployment configuration

## Testing

- `uv run pytest tests/test_capabilities_router.py -q`

This provides the server-authoritative ceiling for Agent Canvas, rather than requiring clients to duplicate the default.


hieptl (approved):
Thank you! 🙏

PR #297, opened 2026-08-01
feat: add Activity Log runs export endpoint

ShashwatXD:
## Summary
- Add `GET /v1/{id}/runs/export` (JSON/CSV) with pagination
- Share run-fetch helpers between list and export
- List + export are pagination-only

## Fixes
OpenHands/OpenHands#16157

## Tests
- Export JSON/CSV across multiple pages
- List runs with `limit`/`offset` only
- Export unit tests pass


neubig (changes requested):
Thanks for the contribution and for adding thorough coverage for Activity Log export. I think we should avoid introducing a separate export endpoint here.

`GET /v1/{automation_id}/runs` already provides the data needed for this feature (`id`, status, timestamps, `conversation_id`, and `error_detail`) with `limit`/`offset` pagination and a total. The client already has the automation name and trigger, can derive the conversation URL from its own origin, calculate duration from timestamps, and generate CSV locally. It can page the existing endpoint to retrieve the complete history (and we can raise the list limit if reducing request count is important).

Please rework this to reuse the existing runs endpoint and remove the new `/runs/export` API and export-specific response types/helpers. A dedicated export API would make sense later for server-side streaming/asynchronous exports or a separately supported external export contract, but neither is needed for the current UI.

ShashwatXD:
we dont need this pr anymore, closing this

Issue #291, opened 2026-07-31
Tracker: automation service hardening — self-inflicted load and false failure signals


aivong-openhands:
Tracking issue for a set of related reliability defects in the automation
service, found while investigating a burst of HTTP 503s on the GitHub webhook
endpoint (`POST /api/automation/v1/events/{uuid}/github`).

## What happened

Readiness probes for the webhook-serving deployment had been timing out
continuously for six days — hundreds of failures — beginning minutes after the
`1.3.1` rollout. Eventually enough instances failed at once that the ejected ones
concentrated their traffic onto the remaining instances, which saturated in turn
and began refusing connections. It was never a full outage: most requests were
served normally throughout, with roughly a third failing over about seven
seconds.

## Root causes

Four independent defects, each filed separately below.

**1. `/ready` is too expensive to be a probe.** It acquires a pooled connection,
`pool_pre_ping=True` issues its own `SELECT 1` on checkout, and the handler then
issues a second one — two database round-trips plus a pool acquisition per probe.
That does not reliably fit inside a short probe timeout at normal managed-Postgres
latency, which is why the probe failed continuously under otherwise healthy
conditions. The second query is also redundant: `pool_pre_ping` has already
proven the connection is live. → #287

**2. Ejecting a slow instance concentrates load onto its neighbours.** Removing
an instance from service redistributes its traffic to the survivors, pushing them
toward the same saturation that got the first one ejected. A probe sensitive
enough to fire under normal load therefore converts a local slowdown into a
spreading one. This is what turned six days of quiet probe failures into a
customer-visible burst. Addressed by making the probe cheap enough that normal
load cannot trip it (#287).

**3. Background loops run once per worker process instead of once per
deployment.** `lifespan` starts the scheduler, dispatcher and watchdog
unconditionally (`app.py:127/143/155`), and `lifespan` runs per uvicorn worker.
Concurrent loop count is `replicas × --workers`, multiplied again by any sibling
deployment running the same ASGI app. The watchdog copies reap the same stale
runs and each issues its own sandbox-API request to verify them, producing enough
request volume to trip the sandbox API's rate limit. It also puts background CPU
on the processes serving inbound webhooks. → #286

**4. A transient sandbox-API error is recorded as a terminal run failure.**
`utils/sandbox.py:69` catches every exception and returns `None`, which is
indistinguishable from "no such sandbox"; the watchdog then marks the run
`FAILED`. So the rate-limit responses caused by (3) killed live automation runs.
[...]

PR #299, opened 2026-08-04
fix: do not mark runs FAILED on transient sandbox API errors

### LINKED ISSUE #285: A transient sandbox-API error (including HTTP 429) marks a running automation as FAILED
## Summary

`get_sandbox_agent_url` treats **any** exception from the sandbox API as proof
the sandbox no longer exists. A rate-limit response, a connect timeout, or a 5xx
all produce the same result as a genuine "no such sandbox", and the watchdog then
marks the customer's running automation as `FAILED`.

`openhands/automation/utils/sandbox.py:69`:

```python
    except Exception as e:
        logger.warning("Failed to get sandbox %s: %s", sandbox_id, e)
        return None
```

`None` is indistinguishable from "not found". The caller turns it into
`error="Sandbox not available"` (`utils/sandbox.py:168`), which the watchdog
records as `FAILED` / `"Timed out: Sandbox not available"`.

## Impact

Dozens of distinct runs were marked `FAILED` in a single 40-minute window
because the sandbox API returned **HTTP 429** to our own polling. Those runs may
well have still been alive.

A rate-limit response carries no information whatsoever about sandbox state.

This is silent: the run reports as timed out, indistinguishable from a genuine
timeout, so neither operators nor users can tell it happened.

The 429s are self-inflicted — see the companion issue about background loops
starting once per uvicorn worker, which multiplies the number of pollers hitting
the sandbox API for the same set of runs.

## Why this blocks failure-count circuit breaking

Any feature that pauses or disables an automation after N failures will consume
this corrupted signal and disable real customers' automations because of an
infrastructure hiccup. N consecutive false failures is easily reachable during a
rate-limit burst. This should be fixed first.

## Proposed fix

Distinguish "not found" from "couldn't check". Return a tri-state (or raise) so
callers can tell the difference, and on "couldn't check" leave the run alone and
retry on the next tick rather than marking it `FAILED`.

Concretely, at minimum: don't collapse `httpx.HTTPStatusError` for 429/5xx and
transport timeouts into the same `None` that means "absent".

Backoff on 429 is worth adding alongside, but is not sufficient on its own — the
correctness bug is the conflation, not the request rate.

nishant-uxs:
## Problem

`get_sandbox_agent_url` collapsed every sandbox-API exception into `None` ("absent"). The watchdog then marked the run `FAILED` with `Timed out: Sandbox not available`. HTTP 429, timeouts, and 5xx carry no information about sandbox liveness — during the #291 incident this false-failed live customer runs.

## Root cause

Watchdog tick (past `timeout_at`) → `backend.verify_run` → `get_sandbox_agent_url` caught `Exception` and returned `None`, so 429 / timeout / 5xx looked identical to "sandbox gone", then `FAILED`.

Broken invariant: "couldn't check" is not the same as "sandbox gone".

## Solution

Minimal semantics fix (no watchdog redesign / no new retry policy):

1. `SandboxLookupTransientError` for 429, 5xx, and transport failures (404 / empty list still mean absent).
2. `VerificationResult.retryable` for transient lookup failures.
3. Watchdog: if `retryable`, leave `RUNNING` for the next tick.

## Failure-state diagram

```
lookup outcome
├── (url, key)      → verify bash → COMPLETED / FAILED as today
├── None (absent)   → FAILED "Sandbox not available" (unchanged)
└── transient error → leave RUNNING (retry next tick)  ← NEW
```

## Test plan

- [x] `pytest tests/test_sandbox_lookup_semantics.py` — 11 passed
  - empty list / HTTP 404 → absent
  - 429 / 503 / timeout → transient (not absent)
  - successful lookup unchanged
  - watchdog: retryable leaves RUNNING; absent still FAILS
- [x] ruff + pyright on changed files

Fixes #285


aivong-openhands:
@nishant-uxs Thank you for your contribution! Some additional changes were needed and another [PR](https://github.com/OpenHands/automation/pull/331) has been merged. Feel free to look at other issues to work on

PR #301, opened 2026-08-04
fix: normalize SQLite telemetry timestamps

### LINKED ISSUE #293: fix(telemetry): normalize SQLite run timestamps before duration arithmetic
## Summary

In local SQLite deployments, watchdog timeout handling can fail while emitting telemetry because it subtracts a timezone-aware `completed_at` from a timezone-naive `started_at`. The exception occurs before the session commit, so the expired run can remain `RUNNING` and be retried indefinitely.

## Reproduction / observed trace

A stale run where Bash verification returns no output reaches the timeout path, then `_base_properties()` raises:

```text
TypeError: can't subtract offset-naive and offset-aware datetimes
  File "openhands/automation/telemetry.py", line 377
    (run.completed_at - run.started_at).total_seconds() * 1000
```

The relevant watchdog messages are:

```text
Could not find bash command result: No bash output found
Could not verify run status: No bash output found, marking as timed out
Marking run as timed out: ...
```

No Bash output should cause a timeout, not prevent the timeout status from being persisted.

## Root cause

SQLite ignores SQLAlchemy `DateTime(timezone=True)` on round-trip and returns persisted timestamps without `tzinfo`. `started_at` is consequently naive after being loaded from SQLite, while the watchdog assigns `completed_at=utcnow()` (aware UTC). Telemetry calculates duration before the terminal update commits.

`ensure_utc()` already exists in `openhands.automation.utils.time`; #139 applies it to Pydantic/API serialization but not telemetry duration arithmetic.

## Proposed fix

Normalize timestamps before subtraction:

```python
from openhands.automation.utils.time import ensure_utc

if run.started_at and run.completed_at:
    duration_ms = int(
        (ensure_utc(run.completed_at) - ensure_utc(run.started_at)).total_seconds() * 1000
    )
```

Also ensure telemetry exceptions cannot abort lifecycle writes (for example, contain telemetry errors or capture telemetry after the terminal status commits).

## Regression coverage

- Telemetry properties accept naive SQLite-style `started_at` plus aware `completed_at` and produce `duration_ms`.
- A watchdog timeout still commits a terminal status when telemetry fails.

## Environment

- `openhands-automation==1.5.0`
- Local Agent Canvas deployment
- SQLite database

---

Linxiushen:
## Summary

- normalize persisted SQLite timestamps to UTC before telemetry duration arithmetic
- make the complete telemetry capture path best-effort so property, consent, identity, and transport failures cannot abort lifecycle writes
- add regression coverage for mixed naive/aware timestamps and pre-transport telemetry failures

## Testing

- `ruff check openhands/automation/telemetry.py tests/test_telemetry.py`
- `ruff format --check openhands/automation/telemetry.py tests/test_telemetry.py`
- 16 focused telemetry tests passed in an isolated Python 3.12 environment
- the full dependency sync was not available locally because LiteLLM 1.93 requires the MSVC linker on Windows; upstream Linux CI should run the complete suite

Fixes #293


neubig (approved):
This fix is verified in production. The stuck `Graham Daily Workflow Prep` automation run was caused exactly by this bug: the watchdog computed `run.completed_at - run.started_at` where one value was timezone-naive (from SQLite) and the other was aware (in-memory), raising `TypeError` inside `capture_automation_event` *before* `session.commit()` — so the status update was rolled back every 60s and the run stayed RUNNING indefinitely.

Both changes here are correct and necessary:
1. `ensure_utc()` normalization fixes the immediate datetime mismatch.
2. Wrapping the entire capture path in try/except ensures telemetry can never block run-state transitions (the structural gap that turned a telemetry bug into a stuck run).

The fix has been applied to the running automation service and confirmed working — the watchdog no longer crash-loops and the run transitioned to COMPLETED.

Reviewed and approved on behalf of @neubig.

PR #302, opened 2026-08-04
perf: cache SDK version metadata

### LINKED ISSUE #289: Telemetry looks up the SDK version via importlib.metadata on every captured event
## Summary

Every captured telemetry event performs an uncached `importlib.metadata` lookup.

`openhands/automation/telemetry.py:327`, inside `_base_properties`:

```python
        **get_server_version_info(missing_sdk_version="unknown"),
```

which reaches `openhands/automation/utils/version.py:18`:

```python
def get_sdk_version() -> str:
    return importlib.metadata.version(SDK_PACKAGE_NAME)
```

`importlib.metadata.version()` scans `site-packages` dist-info on the filesystem.
There is no caching anywhere in `version.py`.

## Impact

Two problems, one small and one structural:

1. It is **synchronous filesystem I/O executed inside an async handler**, so it
   blocks the event loop rather than yielding. Measured at **~0.17 ms per call**
   locally — modest, but it is on the request path and it is pure waste.
2. The value **cannot change during a process's lifetime**, so recomputing it per
   event is unnecessary by construction.

With 3+ captures per inbound webhook this scales linearly with traffic.

## Proposed fix

Cache it — the value is fixed for the life of the process:

```python
@lru_cache(maxsize=1)
def get_sdk_version() -> str:
    return importlib.metadata.version(SDK_PACKAGE_NAME)
```

`get_server_version_info` reads `__version__` directly, which is already cheap,
so caching the SDK lookup is sufficient.

Linxiushen:
## Summary

- cache the installed OpenHands SDK version for the process lifetime
- avoid repeated synchronous package-metadata filesystem lookups on every telemetry event
- add a regression test that verifies repeated reads perform one metadata lookup

## Testing

- `pytest tests/test_version.py --noconftest -q`
- `ruff check openhands/automation/utils/version.py tests/test_version.py`
- `ruff format --check openhands/automation/utils/version.py tests/test_version.py`

Fixes #289

### DIFF
--- a/openhands/automation/utils/version.py
+++ b/openhands/automation/utils/version.py
@@ -1,6 +1,7 @@
 """Version metadata helpers for the automation service."""
 
 import importlib.metadata
+from functools import lru_cache
 from typing import TypedDict
 
 from openhands.automation import __version__
@@ -14,6 +15,7 @@ class ServerVersionInfo(TypedDict):
     sdk_version: str
 
 
+@lru_cache(maxsize=1)
 def get_sdk_version() -> str:
     return importlib.metadata.version(SDK_PACKAGE_NAME)
 
--- a/tests/test_version.py
+++ b/tests/test_version.py
@@ -1,0 +1,22 @@
+import importlib.metadata
+
+from openhands.automation.utils.version import SDK_PACKAGE_NAME, get_sdk_version
+
+
+def test_get_sdk_version_caches_package_metadata_lookup(monkeypatch):
+    calls = 0
+
+    def package_version(name: str) -> str:
+        nonlocal calls
+        calls += 1
+        assert name == SDK_PACKAGE_NAME
+        return "1.2.3"
+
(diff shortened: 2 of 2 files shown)


Linxiushen:
Addressed in 1b86e28: `get_sdk_version()` is cache-free again, so `/sdk-version` and `/server_info` still surface `PackageNotFoundError` as 503 (`tests/test_health.py` back to 9/9, untouched). The telemetry emitter now memoizes only successful `get_sdk_version()` lookups via `lru_cache` (exceptions are not memoized, so a missing SDK is never pinned; `package_version` is still read live). `tests/test_version.py` is replaced by two tests in `tests/test_telemetry.py`: one metadata lookup across repeated reads, and the not-installed path is not cached. Branch rebased onto current `main`.

Verified: `uv run pytest tests/test_health.py tests/test_telemetry.py` (26 passed), `ruff check` / `ruff format --check`, `pyright` on the touched files — all clean.

PR #303, opened 2026-08-04
perf: remove redundant readiness query

### LINKED ISSUE #287: /ready performs two database round-trips per probe, making it fragile under short probe timeouts
## Summary

`/ready` makes **two** database round-trips per call, plus a pool acquisition:

```python
@app.get("/ready")
async def readiness():
    try:
        async with app.state.engine.connect() as conn:
            await conn.execute(text("SELECT 1"))
        return {"status": "ready"}
```

1. acquire a connection from the pool — bounded by `pool_timeout` (default 30s),
   so this can wait
2. `pool_pre_ping=True` (`db.py`) makes SQLAlchemy issue **its own** `SELECT 1`
   on checkout
3. the handler then issues a second `SELECT 1`

## Impact

A readiness probe is typically given a short timeout — Kubernetes defaults
`timeoutSeconds` to `1`. Two round-trips plus a pool acquisition does not
reliably fit in one second at normal managed-Postgres latency, before any
scheduling jitter.

The observable result is a probe that fails intermittently under completely
normal conditions. We saw hundreds of readiness timeouts over six days, which
eventually removed enough instances from service simultaneously that the
survivors absorbed their traffic and saturated in turn.

Note this is also redundant work: `pool_pre_ping` has already proven the
connection is live by the time the handler's own query runs.

## Worth preserving

The handler's `except` branch returning 503 is the part that actually detects an
unreachable database, and it should stay. That detection is independent of the
probe timeout — a genuinely dead DB returns 503 promptly, and pool exhaustion
blocks for `pool_timeout` and fails regardless. So making the check cheaper does
not weaken it.

## Proposed fix

Make `/ready` cost one round-trip at most. Options, roughly in order of
preference:

1. Rely on `pool_pre_ping` and drop the handler's redundant `SELECT 1` — a
   successful `connect()` already implies a live connection.
2. Cache a recent successful result for a second or two so a burst of probes
   doesn't each hit the database.
3. Track DB health from the connection pool's own state rather than probing
   inline per request.

Deployment-side probe tuning is a separate matter for operators, but the endpoint
should be cheap enough not to require it.

Linxiushen:
## Summary

- rely on the engine's configured `pool_pre_ping` during readiness connection checkout
- remove the handler's redundant second `SELECT 1` database round-trip
- tighten the readiness test double so it fails if the endpoint attempts to execute a query

The endpoint still acquires a database connection and preserves the existing 503 response for connection or pre-ping failures.

## Testing

- `ruff check openhands/automation/app.py tests/test_health.py`
- `ruff format --check openhands/automation/app.py tests/test_health.py`
- full endpoint tests were not available locally because LiteLLM 1.93's Windows build requires the MSVC linker; upstream Linux CI should run them

Fixes #287


dylan-openhands (approved):

hieptl (approved):
Thank you! 🙏

PR #304, opened 2026-08-04
perf: reuse telemetry HTTP client

### LINKED ISSUE #288: Telemetry creates a new httpx.AsyncClient per captured event, forcing a TLS handshake per event
## Summary

`capture_automation_event` constructs a new `httpx.AsyncClient` for every
captured event (`openhands/automation/telemetry.py:455`):

```python
        async with httpx.AsyncClient(timeout=2.0) as client:
            response = await client.post(
                f"{settings.posthog_host.rstrip('/')}{POSTHOG_CAPTURE_PATH}",
                json=payload,
            )
```

A client per call means a new TCP connection and a full TLS handshake per event,
with no connection reuse.

## Impact

`receive_event` captures 3+ events per inbound webhook, so this is 3+ TLS
handshakes per request. TLS handshakes are meaningfully CPU-expensive, and this
runs inline in the webhook request path.

It scales linearly with webhook volume, so it is worst exactly when the service
is busiest — which is when the extra CPU is least available.

It also means telemetry latency is paid on the request path with no pooling to
amortise it. During an incident we observed these POSTs taking ~3.0s against
their own configured 2.0s timeout, because the process wasn't being scheduled
promptly.

Note that the majority of these captures are for events that do no work at all —
`automation_event_ignored` fires a full capture, TLS handshake included, for
webhook payloads the service has explicitly decided to discard.

## Proposed fix

Create one `AsyncClient` at application startup (module-level or on `app.state`)
and reuse it, so connections pool and keep-alive applies.

Two worthwhile follow-ons:

1. Skip the capture entirely for discarded events, or batch them — a full
   round-trip per ignored payload is pure overhead.
2. Consider moving capture off the request path (background task or queue) so
   telemetry can never contribute latency to webhook handling. Pooling alone
   removes most of the cost, so this is an improvement rather than a
   prerequisite.

Linxiushen:
## Summary

- reuse one process-wide `httpx.AsyncClient` for PostHog capture requests so telemetry benefits from connection pooling and keep-alive
- initialize the client during application startup and close/reset it during shutdown
- add regression coverage for multi-event reuse and clean client recreation after shutdown

## Testing

- `ruff check openhands/automation/telemetry.py openhands/automation/app.py tests/test_telemetry.py`
- `ruff format --check openhands/automation/telemetry.py openhands/automation/app.py tests/test_telemetry.py`
- 16 focused telemetry tests passed in an isolated Python 3.12 environment
- full dependency sync was unavailable locally because LiteLLM 1.93's Windows build requires the MSVC linker; upstream Linux CI should run the complete suite

Fixes #288

### DIFF
--- a/openhands/automation/app.py
+++ b/openhands/automation/app.py
@@ -30,6 +30,10 @@ from openhands.automation.middleware import (
 from openhands.automation.preset_router import router as preset_router
 from openhands.automation.router import router
 from openhands.automation.scheduler import scheduler_loop
+from openhands.automation.telemetry import (
+    close_telemetry_http_client,
+    get_telemetry_http_client,
+)
 from openhands.automation.telemetry_router import router as telemetry_router
 from openhands.automation.uploads import router as uploads_router
 from openhands.automation.utils.version import get_sdk_version, get_server_version_info
@@ -65,6 +69,7 @@ async def lifespan(app: FastAPI):
 
     # Create shared httpx client for auth (stored in app.state for DI)
     app.state.http_client = create_http_client()
+    get_telemetry_http_client()
 
     # Create engine and session factory, store in app.state
     engine_result = await create_engine(settings)
@@ -183,6 +188,7 @@ async def lifespan(app: FastAPI):
                 pass
 
     await app.state.http_client.aclose()
+    await close_telemetry_http_client()
     await app.state.engine_result.dispose()
     logger.info("Automations service shut down")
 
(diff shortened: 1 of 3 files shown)


Linxiushen:
Rebased onto current main (the inner try/except in `_capture_automation_event` is dropped since the public `capture_automation_event` wrapper on main now owns error handling) and fixed the pyright error in 74fac63: `_MockAsyncClient` now exposes `is_closed` instead of `closed`, mirroring `httpx.AsyncClient.is_closed`, and the restart test asserts `first_client.is_closed is True`, so the attribute resolves on the annotated type.

Verified locally on 74fac63: `uv run pyright` -> 0 errors; `uv run pre-commit run --files openhands/automation/telemetry.py openhands/automation/app.py tests/test_telemetry.py` -> all hooks pass; `uv run pytest tests/test_telemetry.py` -> 17 passed.

PR #305, opened 2026-08-04
fix: trust ingress forwarded headers

### LINKED ISSUE #245: uvicorn ignores X-Forwarded-Proto behind the ingress so the /automations trailing-slash redirect emits an http Location
## Why

When the service runs behind a TLS-terminating ingress (traefik in our deployment), the trailing-slash redirect for the SPA mount emits an insecure `Location`: `GET https://<app-host>/automations` responds `307` with `Location: http://<app-host>/automations/`. The browser recovers via the ingress's http→https redirect, so the UI still loads, but every cold navigation pays an extra redirect hop through a plaintext `http://` URL — and since that hop is insecure, the browser drops the `Secure` session cookie on it.

Root cause is in the container start command, not the ingress. `containers/Dockerfile:45` launches uvicorn bare:

```
CMD ["ddtrace-run", "uvicorn", "openhands.automation.app:app", "--host", "0.0.0.0", "--port", "8000"]
```

Uvicorn's `--proxy-headers` is on by default, but forwarded headers are only trusted from `--forwarded-allow-ips`, which defaults to `127.0.0.1`. In-cluster the connection comes from the ingress pod's IP, so `X-Forwarded-Proto: https` is ignored, the ASGI scope scheme stays `http`, and Starlette builds the redirect URL from that. The blast radius is any absolute URL derived from the request scheme, not just this one redirect.

Fix: add `--forwarded-allow-ips=*` to the CMD (or `ENV FORWARDED_ALLOW_IPS=*`, which uvicorn reads). That is safe for this image because it is only ever deployed behind an ingress; the events deployment launch should get the same treatment.

## Validation

- Reproduced on a live deployment: a browser network trace of `GET /automations` shows `307` → `http://<app-host>/automations/` → second `307` back to `https` before the `200`; the intermediate `http` request carried no cookies (they are `Secure`).
- The pod env contains no `FORWARDED_ALLOW_IPS`, and the chart sets none, matching the uvicorn default-allow-list explanation.

---
This issue was drafted by an AI agent on behalf of the user.

Linxiushen:
## Summary

- trust ingress proxy headers by default in the service container
- let Uvicorn honor `X-Forwarded-Proto: https` from in-cluster ingress addresses so canonical redirects keep the HTTPS scheme
- use the `FORWARDED_ALLOW_IPS` environment variable so direct deployments can override the default with specific trusted proxy IPs or CIDRs

## Validation

- confirmed Uvicorn resolves `Config.forwarded_allow_ips` to `*` from the container environment
- `git diff --check`

Fixes #245

Disclosure: This change was prepared with AI assistance. I reviewed the configuration and validated Uvicorn's environment handling.

### DIFF
--- a/containers/Dockerfile
+++ b/containers/Dockerfile
@@ -2,6 +2,9 @@ FROM python:3.12-slim-bookworm
 
 WORKDIR /app
 ENV PYTHONPATH='/app'
+# The service image runs behind an ingress. Direct deployments can override
+# this with their trusted proxy IPs or CIDRs.
+ENV FORWARDED_ALLOW_IPS='*'
 
 # Datadog labels
 LABEL com.datadoghq.tags.service="automation"


Linxiushen:
Thanks for the review — both non-blocking points are addressed.

Test coverage: commit c5a5503 adds `tests/test_forwarded_headers.py`. It reads the value Docker stores for the `ENV` instruction (shlex unquoting, so the test sees `*`, not `'*'`), asserts `uvicorn.Config` resolves exactly that variable from the environment with `proxy_headers` enabled, and drives `ProxyHeadersMiddleware` end to end: `X-Forwarded-Proto: https` from an arbitrary peer arrives as `request.url.scheme == "https"`. A parametrized negative case pins the quoting trap you flagged — a trusted-hosts value of `'*'` with the quotes kept is an unmatchable host literal and leaves the scheme at `http`. I checked the assertions actually bite by mutating the Dockerfile: deleting the `ENV` line, and storing `"'*'"`, each fail 3 of the 5 tests.

Trust scope: agreed the summary wording is looser than the code. The ingress pod address is not stable, so the in-cluster restriction is deliberately left to the network layer, and `FORWARDED_ALLOW_IPS` stays overridable per deployment for anyone running the image without an ingress in front. I will reword the summary so it does not read as an in-code restriction.

I also rebased onto current `main`, since the previous head was from early August and its CI runs recorded no substantive jobs.

PR #306, opened 2026-08-04
fix: reconcile model index declarations

### LINKED ISSUE #252: Reconcile model index declarations with the actual DB schema (drift)
## Summary
Several columns are declared `index=True` on the models, but the corresponding indexes do not exist in production. The models and the live schema disagree.

## Details
Confirmed absent in production despite being declared `index=True` on the models:
- `automation_runs.timeout_at`
- `automations.user_id`, `.org_id`, `.enabled`, `.deleted_at`, `.last_polled_at`

Context: #157 documented the broader missing-index problem; #160 added the two polling composite indexes; #250 adds `automation_id`. The declarations above are the leftover drift.

## Why it matters
`alembic revision --autogenerate` diffs the models against the DB, so with this drift it will keep proposing `CREATE INDEX` for all of these — risking accidental creation of unused indexes and making autogenerate output noisy and untrustworthy. The models also no longer reflect the real schema.

## Proposed fix
Reconcile per column. For these, dropping the declaration is the better call rather than creating the index:
- `automation_runs.timeout_at` — redundant with the existing `(status, timeout_at)` composite; nothing filters `timeout_at` alone. **Drop `index=True`.**
- `automations.*` — the table is small enough that these indexes give no measurable benefit (the planner favors a sequential scan at this size, and `enabled` is a low-selectivity boolean). **Drop `index=True`**, unless/until `automations` grows materially (at which point `user_id`/`org_id` for the list endpoint may be worth adding deliberately).
- Add a CI guard that fails when `alembic revision --autogenerate` produces a non-empty diff, so model/DB drift can't silently return.

Refs #157, #250.

Linxiushen:
## Summary

- remove ORM declarations for five low-value `automations` single-column indexes that are absent in production
- remove the redundant `automation_runs.timeout_at` single-column index declaration while retaining the deployed `(status, timeout_at)` composite
- add an idempotent cross-database migration so fresh databases and existing deployments converge on the same index set
- extend SQLite migration coverage to assert the redundant indexes are absent and the required run-history/polling indexes remain

## Testing

- ran `alembic upgrade head` against a fresh SQLite database
- verified resulting indexes:
  - `automations`: none of the five drifted single-column indexes
  - `automation_runs`: `automation_id`, `status`, `(status, created_at)`, `(status, timeout_at)`
- ran Alembic `compare_metadata`; none of the six reconciled indexes appeared in the diff
- `ruff check` and `ruff format --check`
- `python -m py_compile` on changed Python files

Fixes #252

Disclosure: This change was prepared with AI assistance. I reviewed the migration and verified it against a fresh SQLite schema.

### DIFF
--- a/migrations/versions/013_reconcile_model_indexes.py
+++ b/migrations/versions/013_reconcile_model_indexes.py
@@ -1,0 +1,54 @@
+"""Reconcile ORM index declarations with the deployed schema.
+
+Revision ID: 013
+Revises: 012
+Create Date: 2026-08-04
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+
+revision: str = "013"
+down_revision: str = "012"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+AUTOMATION_INDEXES = (
+    ("ix_automations_user_id", ["user_id"]),
+    ("ix_automations_org_id", ["org_id"]),
+    ("ix_automations_enabled", ["enabled"]),
+    ("ix_automations_deleted_at", ["deleted_at"]),
+    ("ix_automations_last_polled_at", ["last_polled_at"]),
+)
+RUN_TIMEOUT_INDEX = ("ix_automation_runs_timeout_at", ["timeout_at"])
+
+
+def upgrade() -> None:
+    for name, _columns in AUTOMATION_INDEXES:
+        op.drop_index(name, table_name="automations", if_exists=True)
+
+    op.drop_index(
+        RUN_TIMEOUT_INDEX[0],
+        table_name="automation_runs",
+        if_exists=True,
+    )
+
+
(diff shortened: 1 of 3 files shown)


Linxiushen:
Rebased onto current `main` (1e72055) and renumbered the migration in fa53716. The head on `main` is now `027` (026_add_state_and_trigger_source and 027_add_automation_drafts landed after the review), so the file is `028_reconcile_model_indexes.py` with `revision = "028"` / `down_revision = "027"`.

Verified on the rebased branch: `alembic heads` → `028 (head)` only; `uv run alembic upgrade head` on a fresh SQLite db succeeds with the six reconciled indexes absent; `upgrade 027 → head → downgrade 027` round-trips (indexes present → dropped → recreated); `compare_metadata` shows none of the six; `pytest tests/test_db.py` 29 passed; ruff check/format clean. No autogenerate CI guard added.

PR #313, opened 2026-08-06
feat: add generic preset metadata field to Automation model

### LINKED ISSUE #56: feat: Add generic preset metadata field to Automation model
## Architecture-aligned scope

The metadata goal and existing implementation in #89 remain valid. Extend the work to preserve opaque provenance from an extension-owned automation template without moving template definitions into the automation service.

## Acceptance criteria

- Persist and return an opaque extension template identifier, version, and non-secret submitted configuration for preset-created automations.
- Rebase #89 and keep the schema and migration compatible with both SQLite and PostgreSQL.

## Dependencies

- Blocked by OpenHands/extensions#420.

---

## Summary

Follows #51. Once repository support lands for preset endpoints, we should record the repository configuration (and other preset-specific data) in a generic metadata field that the UI can consume.

## Background

Preset automations (`/v1/preset/prompt` and `/v1/preset/plugin`) are built from user-provided configuration that gets packaged into a tarball. Currently, only the `prompt` field is stored on the `Automation` model (see `models.py:56`). The generated tarball contains additional configuration that would be useful for the UI to display and for users to understand what their automation does:

1. **Repositories** (from #51) — cloned into the sandbox at runtime
2. **Plugins** — the `PluginSource` list used by `/v1/preset/plugin` (stored in `plugins_config.json` inside tarball)
3. **Prompt** — already stored as a top-level field, but could be included in metadata for consistency

## Proposal

Add a `preset_metadata` JSONB column to the `Automation` model:

```python
# Preset-specific metadata (populated by preset endpoints, NULL for custom SDK automations)
preset_metadata: Mapped[dict | None] = mapped_column(JSONB, nullable=True)
```

### Metadata Schema

The field is intentionally generic to accommodate future presets:

```json
{
  "preset_type": "prompt" | "plugin",
  "prompt": "...",
  "plugins": [{"source": "github:owner/repo", "ref": "v1.0.0"}],
  "repos": [{"url": "OpenHands/repo", "ref": "main"}]
}
```

For prompt presets: `preset_type`, `prompt`, `repos` (once #51 lands)
For plugin presets: `preset_type`, `prompt`, `plugins`, `repos`

### Changes Required

1. **`models.py`** — Add `preset_metadata` JSONB column
2. **`schemas.py`** — Add `preset_metadata: dict | None` to `AutomationResponse`
3. **`preset_router.py`** — Populate `preset_metadata` when creating automations:
   - Prompt preset: `{"preset_type": "prompt", "prompt": body.prompt, "repos": [...]}`
   - Plugin preset: `{"preset_type": "plugin", "prompt": body.prompt, "plugins": [...], "repos": [...]}`
4. **Migration** — Add the new column

### Note

This does not apply to custom SDK automations (created via `/v1` with a user-uploaded tarball), which will have `preset_metadata = NULL`.

---

*This issue was created by an AI assistant (OpenHands).*

hieptl:
### Why

Preset automations (`POST /v1/preset/prompt`, `POST /v1/preset/plugin`) are built from user-provided configuration — a prompt, plugin sources — that gets baked into a generated tarball. Only the `prompt` survives as a column on `Automation`; everything else is opaque inside the tarball. The extension-owned Automation UI needs to display what an automation was built from, and today it can't.

### Summary

- **`Automation.preset_metadata`** — new nullable JSON column recording the creation-time preset configuration. Uses generic `sa.JSON`, not the JSONB proposed in #56: `AGENTS.md` forbids PostgreSQL-only types because self-hosted deployments run SQLite, both existing JSON columns already use `sa.JSON`, and this field is display metadata that is never queried or indexed.
- **Preset endpoints populate it** at creation:
  - prompt preset → `{"preset_type": "prompt", "prompt": ..., "repos": [...]}` (`repos` only when provided)
  - plugin preset → `{"preset_type": "plugin", "prompt": ..., "plugins": [...], "repos": [...]}` (each list only when provided; variant-based A/B automations omit `plugins`, matching the issue's schema which doesn't cover variants)
  - repos/plugins serialized with `model_dump(exclude_none=True)`, exactly as the tarball configs (`repos_config.json` / `plugins_config.json`) already are — which also means `PluginSource`'s credential-redacting serializer applies.
- **`AutomationResponse.preset_metadata`** — exposed on the API response (`None` for non-preset automations).
- **`PATCH /v1/{automation_id}`** — when a prompt edit regenerates the preset tarball, `preset_metadata["prompt"]` is updated too, so the metadata never contradicts what the automation actually runs. The dict is reassigned rather than mutated because in-place mutation of a JSON column is not change-tracked by SQLAlchemy.
- **Migration `013_add_preset_metadata.py`** — adds the column; downgrade drops it.
- Custom SDK automations (`POST /v1` with a user-uploaded tarball) are untouched and keep `preset_metadata = NULL`. Preset automations created before this change also have `NULL` — there is no backfill; metadata is recorded at creation time only.

### Issue Number

Resolves https://github.com/OpenHands/automation/issues/56

### How to Test

**1. Automated checks**

```bash
uv run ruff check && uv run ruff format --check .
uv run python -m pytest tests/ -q     # 1087 passed (6 new)
```

**2. The new tests bite.** With the source changes stashed and only the tests applied, all six fail; with the implementation restored, all pass:

- `test_create_from_prompt_stores_preset_metadata` — prompt preset, no repos → metadata is exactly `{preset_type, prompt}` (no `repos` key)
- `test_create_from_prompt_stores_repos_in_preset_metadata` — repos serialized with `exclude_none` (no `ref` key on a repo without one)
- `test_create_from_plugin_stores_preset_metadata` — plugins + repos recorded
- `test_ab_automation_preset_metadata_omits_plugins` — variant-based automation → no `plugins` key
- `test_create_automation_preset_metadata_is_null` — raw `POST /v1` → `preset_metadata: null`
- `test_update_prompt_syncs_preset_metadata` — PATCH prompt edit updates the metadata prompt and preserves the other keys

**3. Migration cycle** against a scratch SQLite database:

```bash
AUTOMATION_DB_URL=sqlite:////tmp/scratch.db uv run alembic upgrade head   # → 013 (head)
sqlite3 /tmp/scratch.db "PRAGMA table_info(automations);" | grep preset   # preset_metadata|JSON|nullable
AUTOMATION_DB_URL=sqlite:////tmp/scratch.db uv run alembic downgrade 012  # column dropped cleanly
```

**4. Manual sanity** — create a preset automation and read it back:

```bash
curl -X POST .../api/automation/v1/preset/prompt -d '{
  "name": "Daily PR summary",
  "prompt": "Summarize open PRs",
  "trigger": {"type": "cron", "schedule": "0 9 * * 1"},
  "repos": [{"url": "owner/repo", "provider": "github"}]
}'
# response includes:
# "preset_metadata": {"preset_type": "prompt", "prompt": "Summarize open PRs",
#                     "repos": [{"url": "owner/repo", "provider": "github"}]}
```

A raw `POST /v1` automation returns `"preset_metadata": null`.

### DIFF
--- a/migrations/versions/013_add_preset_metadata.py
+++ b/migrations/versions/013_add_preset_metadata.py
@@ -1,0 +1,30 @@
+"""Add preset_metadata column to automations table.
+
+This migration adds a nullable preset_metadata JSON column to the automations
+table so preset endpoints can record the configuration used to build the
+automation (preset type, prompt, plugins, repos) for the UI to consume.
+Custom SDK automations leave it NULL.
+
+Revision ID: 013
+Revises: 012
+Create Date: 2026-08-01
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+
+revision: str = "013"
+down_revision: str = "012"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column("automations", sa.Column("preset_metadata", sa.JSON, nullable=True))
+
+
+def downgrade() -> None:
+    op.drop_column("automations", "preset_metadata")
--- a/migrations/versions/014_add_preset_metadata.py
+++ b/migrations/versions/014_add_preset_metadata.py
@@ -5,8 +5,8 @@ table so preset endpoints can record the configuration used to build the
 automation (preset type, prompt, plugins, repos) for the UI to consume.
 Custom SDK automations leave it NULL.
 
-Revision ID: 013
-Revises: 012
+Revision ID: 014
+Revises: 013
 Create Date: 2026-08-01
 """
 
@@ -16,8 +16,8 @@ import sqlalchemy as sa
 from alembic import op
 
 
-revision: str = "013"
-down_revision: str = "012"
+revision: str = "014"
+down_revision: str = "013"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
--- a/openhands/automation/models.py
+++ b/openhands/automation/models.py
@@ -61,6 +61,11 @@ class Automation(Base):
     # Optional prompt (set when created via preset endpoints)
     prompt: Mapped[str | None] = mapped_column(Text, nullable=True)
 
+    # Preset-specific metadata (populated by preset endpoints; NULL for custom
+    # SDK automations).
+    # Uses generic JSON type for cross-database compatibility (PostgreSQL + SQLite)
+    preset_metadata: Mapped[dict | None] = mapped_column(JSON, nullable=True)
+
     # Model profile name to use for automation runs.
     # None is only used for legacy/local fallback.
     model: Mapped[str | None] = mapped_column(String(64), nullable=True)
--- a/openhands/automation/preset_router.py
+++ b/openhands/automation/preset_router.py
@@ -454,12 +454,20 @@ async def create_automation_from_prompt(
     # 3. Create the automation referencing the internal upload
     tarball_path = build_internal_url(upload_id)
 
+    preset_metadata: dict[str, Any] = {
+        "preset_type": "prompt",
+        "prompt": body.prompt,
+    }
+    if body.repos:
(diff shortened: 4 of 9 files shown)


tofarr (approved):
🍰

PR #327, opened 2026-08-11
feat: sync automations to a git repository

### LINKED ISSUE #300: Add 'sync to git' functionality for automations
## Summary

Add a "sync to git" feature where all automations on a particular VM / agent server are synced to a git repository (or a directory within a repo). When automations change on the VM or in git, they should auto-update in the other direction.

## Motivation

Right now on a VM it's hard to see both:
1. **The current state** of the automations
2. **How they have changed over time**

Syncing automations to git would provide:
- **Visibility** — a readable, reviewable snapshot of all automations in a repo
- **Controllability** — edit automations via PRs, reviews, and standard git workflows
- **Versioning** — full history of how automations evolved, with the ability to diff/rollback

This is similar to Overleaf's git sync functionality, which has proven handy.

A related use case: **backups for canvas configuration**. Once everything is set up, it's painful to redo it all if a VM needs to be rebuilt. Git sync would serve as an automatic backup.

## Proposed Design

Based on discussion:

- The user (i.e. the operator of an agent server) should define the git repo, or even a directory within a repo.
- One setting per **agent server**: a single agent server has a single list of automations and a single GitHub repo (or directory) to sync to.
- The singular backend (the active agent-server) would have a git repo where it syncs automations to.

### Open Questions

- Should this be scoped to automations only, or also cover other canvas-wide configuration (e.g. extensions like skills + plugins)?
- Initial proposal: user defines the repo individually per agent server. Could later expand to a canvas-wide git syncing location for other things like extensions.

## Origin

Discussion in Slack (Graham Neubig, Devin Vinson).

---

_This issue was created by an AI agent (OpenHands) on behalf of Graham Neubig._

VascoSch92:
> This repo has no PR template, so this follows the structure of the one in
> `OpenHands/OpenHands` for consistency.

## Why

Closes #300 — automations live only in a VM's database today, so there's no
readable snapshot of what exists, no history of how it changed, and no backup
if the VM has to be rebuilt.

## Summary

- Serializes each automation to `{path}/{slug}/` as `automation.yaml` plus its
  extracted tarball contents, and syncs it with a git repo in both directions:
  automations edited through the API are pushed, and changes made in git (e.g.
  via a PR) are imported back.
- Adds `GET /v1/git-sync/status`, `PUT /v1/git-sync/config` and
  `POST /v1/git-sync/sync`. Repo, branch, path, credentials, commit author and
  sync interval are all reconfigurable at runtime without a restart.
- **Sync is manual by default** — see the deviation from the issue below.
- Optional Fernet encryption of file contents before they are committed.

## Issue Number

OpenHands/automation#300

## How to Test

Run the service against a throwaway repo — a local bare repo works and needs no
credentials:

```bash
git init --bare -b main /tmp/git-sync-remote.git

AUTOMATION_GIT_SYNC_ENABLED=1 \
AUTOMATION_GIT_SYNC_REPO_URL=/tmp/git-sync-remote.git \
AUTOMATION_AGENT_SERVER_URL=http://localhost:3000 \
uv run uvicorn openhands.automation.app:app --port 18001
```

Then, with `$KEY` set to your session API key:

```bash
# 1. status — manual-only by default
curl -H "X-Session-API-Key: $KEY" localhost:18001/api/automation/v1/git-sync/status

# 2. create an automation, then sync it
curl -X POST -H "X-Session-API-Key: $KEY" localhost:18001/api/automation/v1/git-sync/sync

# 3. the automation is now committed and pushed
git -C /tmp/git-sync-remote.git ls-tree -r HEAD --name-only
```

Verified end-to-end against a real GitHub repo, not just the unit tests:

- Created an automation, triggered a sync, and confirmed the commit landed on
  the remote containing `automations/<slug>/automation.yaml` plus
  `tarball/{main.py,prompt.txt,setup.sh}`.
- Changed the repo URL at runtime from the local bare repo to a GitHub repo and
  confirmed the existing commit was carried over to the new remote.
- Changed `path` at runtime and confirmed every automation was re-exported
  under the new directory, with the old one left in place and a warning logged.
- Set `interval_seconds` to 20 on the running service and confirmed automatic
  syncing began ~40s later with no restart, then set it back to 0 and confirmed
  `last_synced_at` stopped advancing over 75s.
- `uv run pytest` — 1212 passing.
- `pre-commit run --from-ref main --to-ref HEAD` — ruff format, ruff lint,
  pycodestyle and pyright all clean.

## Type

- [x] Feature
- [x] Bug fix

## Notes

**Deviation from the issue.** #300 specifies automatic sync ("they should
auto-update in the other direction", `git_sync_interval_seconds: int = 60`).
This ships **manual by default** (interval `0`): a cycle is bidirectional,
resolves conflicts by overwriting the losing side, and propagates deletions
both ways, so running it unattended on a timer can rewrite the repo or the
database with no one watching. Automatic sync is one field away — set an
interval in the UI — but it is opt-in. Happy to flip the default if you'd
rather match the issue as written.

**Commit layout.** Seven commits, each green in isolation:

- `8c5ef41` is an independent pre-existing bug on `main` (the migrations
  directory could not be found when running from a source checkout) and can be
  cherry-picked on its own.
- `e27adc2` is the feature as first written, including two bugs.
- `e2c22d1`, `45542dc` and `f57587b` fix those and others found while testing
  live. They are kept separate because the failure mode is worth reading: all
  of them made a cycle **report success while doing nothing** — no exception,
  so no error recorded, so `/status` showed a healthy sync while automations
  stayed on local disk. Each fix ships a regression test that was verified to
  fail without it.

**Not env-configurable.** `interval_seconds` is the one git-sync setting with
no environment variable — it is runtime-only, set from the UI, so there is a
single place to change it.

**Conflict policy** is last-writer-wins per whole automation, as proposed in
the issue. There is no merging: the losing edit is overwritten and recoverable
only from git history. Branch-level divergence stops sync with an error rather
than force-pushing.

**UI:** OpenHands/OpenHands#16521 adds the Git Sync page that drives these
endpoints. The API is fully functional without it.


openhands/automation/config.py
```diff
@@ -284,6 +285,111 @@ def enabled(self) -> bool:
...
+    """Git sync configuration for backing up/versioning automations in git.
+
+    When enabled, automations are serialized to files and pushed to a git
+    repo, and changes pushed there (e.g. via a PR) are pulled back. Only
+    active in local/self-hosted mode -- a single repo maps to a single
+    agent server, which doesn't fit the multi-tenant SaaS deployment.
+
+    Environment variables (AUTOMATION_ prefix):
+        AUTOMATION_GIT_SYNC_ENABLED: Enable git sync (default: False). Only
+            takes effect in local mode; ignored (with a warning) otherwise.
```
neubig: I don't think that we need a feature flag for this? If it's not configured it's not on, so for simplicity we can probably just remove this.

VascoSch92: Oh yes... it is a leftover actually

neubig (approved):
I have been using this and it works well.

PR #329, opened 2026-08-12
feat: persist automation task outcomes

malhotra5:
## Summary
- Pin automation to the SDK PR commit that reports task outcomes.
- Add `AutomationRun.task_outcome` storage, migration, callback schema/handler support for `agent_outcome` and `task_outcome`, and response serialization.
- Advertise callback support to SDK scripts and recover task outcomes through agent-server conversation metadata when watchdog fallback verifies a missed callback.
- Print conversation IDs from prompt/plugin presets so fallback recovery can locate the conversation reliably.

## Tests
- `uv run pre-commit run --files openhands/automation/models.py openhands/automation/schemas.py openhands/automation/router.py openhands/automation/watchdog.py openhands/automation/utils/agent_server.py openhands/automation/utils/sandbox.py openhands/automation/dispatcher.py openhands/automation/execution.py openhands/automation/presets/prompt/sdk_main.py openhands/automation/presets/plugin/sdk_main.py tests/test_router.py tests/test_watchdog.py tests/test_local_mode.py tests/test_db.py tests/test_schemas.py migrations/versions/015_add_run_task_outcome.py --show-diff-on-failure`
- `uv run pytest tests/ -q --ignore=tests/integration`

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/2b8666dc-2f64-4f22-90b2-0945f643ca51)


smolpaws (commented):
🐾 smolpaws (for Engel). The watchdog fallback here (`get_conversation_task_outcome` → `GET /api/conversations/:id` → `.task_outcome`) can read a **stale** outcome.

`task_outcome` is persisted on the conversation and never reset on resume (SDK #4465). So a conversation that continues past its first `finish` — e.g. a Slack run that replies, user messages again, same conversation resumes — keeps the previous run's outcome until a new one is written. If that resumed run ends without a `finish`, the fallback recovers the *previous* run's outcome.

Best fixed on the SDK side (clear/mark in-progress at run start); flagging here so the recovery path is aware. Primary callback path is fine when a run sets a fresh outcome.

jpshackelford:
@malhotra5 @DevinVinson 

I'd like to propose an alternative. One of the challenges of using automations is that when they scale up status of single runs is not important nor does it turn out that one automation equals one use case. Often you will have a network of automations or other conversations spawned by automations contributing to a use case result. 

I think what we need is the concept of an activity log. A user can create a named automation activity log for a given use case that can receive arbitrary json(for agents) and markdown (for humans) entries and the automation prompt can instruct the agent to write things to the log that are important to give an overview of what is happening in the system. Then we display the log last entry first and you can quickly understand what is happening by scanning the stream. That is a much stronger signal than I would get from metadata associated with a single automation run result or even a series of run result that is tightly coupled to a single automation.

Issue #330, opened 2026-08-12
Register automation run conversation ID as soon as conversation is created


malhotra5:
## Problem

Automation runs currently do not expose their `conversation_id` until the run completion callback fires. This means a long-running automation can be marked `RUNNING` in the automation UI/API while still having no linked conversation, even though the SDK has already created the OpenHands conversation.

This is a poor UX because users cannot open or inspect the live conversation until the automation finishes or fails.

## Current behavior

From the current automation and SDK flow:

- The dispatcher marks an `AutomationRun` as `RUNNING` and injects `AUTOMATION_CALLBACK_URL` plus `AUTOMATION_RUN_ID` into the sandbox.
- The preset script creates a `Conversation(...)` during execution.
- In the software agent SDK, `RemoteConversation` calls `workspace.register_conversation(str(self._id))` immediately after the agent-server conversation is successfully created.
- `register_conversation()` currently only stores the ID in memory.
- The automation service stores `AutomationRun.conversation_id` only later, when `POST /v1/runs/{run_id}/complete` receives the completion payload.

Relevant automation files:

- `openhands/automation/models.py` — `AutomationRun.conversation_id`
- `openhands/automation/router.py` — `/v1/runs/{run_id}/complete`
- `openhands/automation/dispatcher.py` — callback env var injection
- `openhands/automation/presets/prompt/sdk_main.py` and `openhands/automation/presets/plugin/sdk_main.py` — conversation creation

Relevant SDK behavior:

- `RemoteConversation` already knows the conversation ID immediately after `POST /api/conversations` succeeds.
- `RemoteWorkspace.register_conversation()` and `OpenHandsCloudWorkspace.register_conversation()` are the natural hook points for early registration.

## Desired behavior

As soon as the SDK successfully creates the conversation, the automation run should be updated with the conversation ID. The existing completion callback should continue to send `conversation_id` as a backward-compatible fallback.

## Possible implementation

One possible design:

1. Add a non-terminal automation endpoint, for example:

   ```http
   POST /api/automation/v1/runs/{run_id}/conversation
   ```

   Request body:

   ```json
   {"conversation_id": "..."}
   ```

   This endpoint should authenticate like the completion callback, verify ownership, update only `AutomationRun.conversation_id`, and not change run status, completion time, cost, or sandbox cleanup behavior.

2. Inject a new env var during dispatch, for example:

   ```text
[...]

PR #331, opened 2026-08-12
fix: stop marking successful automation runs as FAILED

hieptl:
- [x] A human has tested these changes.

---

## Why

Automation runs are marked **Failed** even though the agent conversation they spawned completed successfully (Linear OHE-2822: reproduced on a Replicated stable environment, Joe's "Weekly Replicated Report" failing every run since Jul 13, Modern Health's webhook automation marked FAILED "right after" the conversation completes, with an identical error on every run).

Root cause — the run outcome is derived from the wrapper process's fate under two skewed clocks, never from the conversation's outcome:

1. **Deadline skew.** `timeout_at` starts counting at dispatch (`dispatch_pending_runs` → `mark_run_status`), but the bash command running `setup.sh` + `sdk_main.py` gets the _same_ budget enforced by the agent-server **from bash start** — after sandbox provisioning (up to 300s), tarball upload, and the SDK install (minutes). The watchdog deadline therefore always expires before the in-sandbox budget does.
2. **"Still working" recorded as failure.** At `timeout_at`, verification reports a healthy in-flight command as `"Command still running"` — or, more commonly, `"No bash output found"`, because the agent-server only emits `BashOutput` events on 1 MiB stream chunks or on completion. The watchdog's unverified path then **deleted the sandbox** (unless `keep_alive`) and marked the run `FAILED "Timed out: …"`. The conversation — an autonomous server-side loop on the sandbox's agent-server — kept running and typically finished successfully.
3. **Late success discarded.** `complete_run` only updates `WHERE status == RUNNING`, so a COMPLETED callback landing after the watchdog's decision got a 409 and was thrown away.
4. **In-sandbox tail in the FAILED blast radius (secondary).** After a successful `conversation.run()`, the preset tail (settle loop, live stats fetch, `close()`) still ran inside the workspace context; any exception there made the SDK's `__exit__` callback report FAILED for a successful conversation.

## Summary

- **Two-phase watchdog deadline.** At dispatch, `timeout_at` is padded with the provisioning budget (`sandbox_ready_timeout + run budget + margin`) so runs that die during provisioning are still reaped. Once `execute_in_context` succeeds, the new RUNNING-guarded `update_run_timeout_at` helper re-anchors it to `bash start + run budget + margin` — the bash service's own kill now always fires first, so watchdog verification finds a concrete exit code (0 → COMPLETED via the existing missed-callback rescue; -1 → genuine timeout with real stderr).
- **Bounded deferral instead of destruction.** The two still-running verification results (`STILL_RUNNING_VERIFICATION_ERRORS`) defer `timeout_at` by one watchdog interval per scan — no FAILED, no sandbox deletion — bounded by `started_at + sandbox_ready_timeout + run budget + hard grace`. Past the cap, behavior is exactly as before (cleanup honoring `keep_alive`, `Timed out: …`, telemetry, first-run outcome). `mark_stale_runs` now commits unconditionally so deferrals persist. The `"Sandbox not available"` and verification-exception paths are deliberately untouched (transient sandbox-API errors are #285's scope).
- **Late-success reconciliation.** A COMPLETED callback that meets a FAILED row whose `error_detail` starts with `"Timed out: "` (written only by the watchdog) flips the run to COMPLETED — clearing `error_detail`, storing `conversation_id`/`cost`, and emitting `automation_run_completed` with `reconciled_watchdog_timeout: true`. Late FAILED callbacks, dispatcher-authored failures, and cancelled runs still 409.
- **Preset tail hardening** (both `presets/{prompt,plugin}/sdk_main.py`): the settle loop + stats fetch and the `close()` in `finally` are now non-fatal; `conversation.run()` exceptions still propagate. The callback's `cost` is registered by `close()` from cached stats, so the payload is unaffected. Tarballs are frozen per automation, so this part reaches new/edited automations — the service-side fixes above carry the existing fleet.
- New `SandboxSettings` knobs: `run_timeout_margin` (120s, env `AUTOMATION_RUN_TIMEOUT_MARGIN`) and `run_timeout_hard_grace` (600s, env `AUTOMATION_RUN_TIMEOUT_HARD_GRACE`). No DB schema changes — `timeout_at` already exists; note it is API-visible and now moves at bash start and during deferrals, consistent with its documented "deadline" meaning.

Known residual (called out deliberately): a conversation finishing within seconds of the bash kill can still lose its callback and be marked FAILED at the watchdog with exit -1 — the window shrank from unbounded to roughly the margin; closing it fully would require conversation-outcome-based decisions, kept out of this minimal fix. A wedged-but-reachable sandbox now lives up to `hard_grace` longer before cleanup — a deliberate trade against destroying live conversations.

## Issue Number

N/A.

Related, intentionally out of scope: #285 (transient sandbox-API errors marked as terminal failures), #330 (register conversation_id early).

## How to Test

```bash
uv run python -m pytest tests/test_watchdog.py tests/test_dispatcher.py tests/test_router.py -v
make test        # full suite: 1144 passed (needs Docker for testcontainers)
make lint && make format

```

New tests (10, written to fail on the pre-fix code — verified by stashing the source changes):

- `tests/test_watchdog.py::TestVerifyAndMarkRunStillRunning` — both still-running strings defer `timeout_at` (≤ one watchdog interval) with no FAILED and no cleanup; exhausted hard grace falls through to the terminal timeout path (FAILED + cleanup); a callback that lands mid-scan wins the race and the deferral is a no-op.
- `tests/test_dispatcher.py::TestEffectiveTimeout::test_successful_dispatch_resets_timeout_to_bash_start` — after a successful dispatch, `timeout_at` is re-anchored to bash start + budget + margin (provisioning padding dropped).
- `tests/test_dispatcher.py::TestUpdateRunTimeoutAt` — the RUNNING guard never resurrects a terminal run's deadline.
- `tests/test_router.py::TestCompleteRun` — late COMPLETED over a watchdog-authored `"Timed out: "` FAILED reconciles (status, cleared `error_detail`, `conversation_id`, `cost`); late FAILED, dispatcher-authored FAILED (`"Internal error"`), and CANCELLED all still 409.

Updated: the two `TestEffectiveTimeout` assertions now expect the phase-1 provisioning-padded deadline (computed from `get_config().sandbox`, not literals).

Note: `test_create_automation_success` and `test_dispatch_automation_success` fail on a clean `main` checkout in this environment (`telemetry_distinct_id` is None) — pre-existing and unrelated to this change.

Manual trace: dispatch an automation whose conversation outlives the old deadline — the run now stays RUNNING through watchdog scans (deferral log lines), then flips COMPLETED on the callback; and a run already FAILED `"Timed out: …"` flips to COMPLETED when the late callback arrives.

## Video/Screenshots

N/A — backend-only change; behavior is pinned by the tests above.


tofarr (approved):
🍰

PR #335, opened 2026-08-12
chore: add success logging for tarball storage writes and deletes

jpshackelford:
## Summary

Adds INFO-level audit logging on the **success path** of the automation object store. Today the storage layer is silent when a write or delete succeeds and only logs on error, so there is no positive record of *when* a tarball object was created or removed. This change makes those lifecycle events observable.

Changes:
- `S3FileStore.write`, `write_stream`, and `delete` now emit a success log line with `bucket`, `path`, and (for writes) `size_bytes`, plus structured `extra` fields for querying in log aggregation.
- `regenerate_preset_prompt_tarball` now logs the `automation_id`, the old/new upload IDs and storage paths, and whether the superseded object was actually removed — so a delete can be attributed to a specific automation and prompt-edit request.

No behavior changes; logging only.

## Motivation — the failure scenario this is meant to catch

We are investigating a customer-reported failure (customer not named here) where triggering an automation fails almost immediately with a generic "Internal error." The proximate cause is confirmed from logs: the dispatcher tries to read the automation's stored tarball and the object is **missing** from the bucket (`NoSuchKey` → `FileNotFoundError`), even though the corresponding upload record still exists in the database and is marked `COMPLETED`. The automation's `tarball_path` therefore references an object that is no longer physically present in storage.

What we could **not** determine from the available logs is *how the object went missing*, because the two candidate mechanisms leave no trace today:

1. **The application deleted it.** When a preset automation's prompt is edited, `regenerate_preset_prompt_tarball` deletes the superseded object. The successful delete currently logs nothing, so we cannot see whether (or when) the app removed a given object, nor tie it to the edit that triggered it.
2. **The object store lost it** underneath a live record (e.g. an interrupted write / storage-side data loss). The app never issues a delete in this case, so again there is no log entry — and no successful-write entry to prove the object ever landed.

With this logging in place, the two cases become distinguishable in the logs:
- A `regenerate preset tarball` / `S3 delete ok` line for the missing object's path ⇒ the application removed it (case 1).
- An `S3 write ok` / `write_stream ok` for the path but **no** corresponding delete ⇒ the object was written successfully and later disappeared without the app deleting it (case 2).

This is intended to help isolate the root trigger of the reported failure going forward and, more generally, to give us a lifecycle trail for automation tarballs (created / superseded / deleted) that survives in log aggregation.

## Testing

- `uv run pytest tests/test_storage_s3.py tests/test_storage.py tests/test_preset_router.py` → 136 passed, 41 skipped (skipped are the real-S3 integration tests).
- `uv run ruff check` and `ruff format --check` pass on both changed files.

---
_This pull request was created by an AI agent (OpenHands) on behalf of the OpenHands team._

@jpshackelford can click here to [continue refining the PR](https://app.all-hands.dev/conversations/cecb97de-11ae-47a3-a14a-953e0dd95897)


juanmichelini (commented):
Superseded.

juanmichelini (commented):
Superseded.

juanmichelini:
Suggestion: split the preset tarball delete audit into two fields so logs distinguish an app-issued delete from an object that was already missing.

```python
old_object_delete_succeeded = False
old_object_already_missing = False

try:
    file_store.delete(source_upload.storage_path)
    old_object_delete_succeeded = True
except FileNotFoundError:
    old_object_already_missing = True
except Exception as e:
    logger.exception(
        "Failed to delete superseded tarball at %s: %s",
        source_upload.storage_path,
        e,
    )

file_removed = old_object_delete_succeeded or old_object_already_missing
```

Then log `old_object_delete_succeeded` and `old_object_already_missing` separately instead of a single `old_object_removed` value. This preserves the distinction the audit trail is meant to support.

jpshackelford:
Agreed, and thanks for catching this. Your two-field split is the right call: **old_object_already_missing** is itself a symptom of the failure we are investigating (a supersede that finds nothing to delete), so folding it into the same boolean as a normal successful delete would discard exactly the signal this audit trail is meant to surface.

This was implemented in 3c97421 and then accidentally reverted in 5d8df1b (revert intended as a comment, per the note above). I have reapplied it in 57ee395, so the branch now logs old_object_delete_succeeded and old_object_already_missing separately. Lint/format clean and preset_router tests pass (96 passed, 41 skipped).

_This comment was posted by an AI agent (OpenHands) on behalf of the OpenHands team._

juanmichelini (approved):
LGTM!

PR #340, opened 2026-08-14
chore: remove QA changes workflow

neubig:
## Summary
- remove the dedicated QA Changes by OpenHands workflow
- stop the broken `qa-changes` action from running on labeled pull requests

## Validation
- `git diff --check`
- confirmed no remaining QA workflow references under `.github/workflows`

This PR was created by an AI agent (OpenHands) on behalf of the user.


enyst (approved):
SGTM, I do think we’re currently missing an automated QA though. I agree this workflow wasn’t all that good, I don’t know if Vasco has an Automation doing the main QA (before/after) on PRs though, if not I think I’ll propose something like that.

PR #346, opened 2026-08-18
feat: accept catalog bundle automations on the raw create path

### LINKED ISSUE #344: Support catalog-shipped bundle automations: template provenance, preflight, and a capability flag on the raw create path
## Context

Catalog entries in OpenHands/extensions currently ship a prompt: the host renders a setup form and posts the result to `POST /v1/preset/prompt`, and the service builds the tarball. We want an entry to be able to ship **a script tarball instead** — a "bundle" — for the automations that are deterministic machinery rather than judgement: polling, dedupe, state, fixed API calls. `github-pr-reviewer` is the first case; its script already does its own polling, per-repo KV state and checkout lifecycle, and starts a conversation only when a label event actually needs a review.

**The service needs no new endpoint for this.** The host packs the entry's bundle files plus a rendered `config.json`, then:

1. `POST /v1/uploads?name=<entry-id>` with `Content-Type: application/gzip` → `oh-internal://uploads/<uuid>`
2. `POST /v1` with `name`, `trigger`, `tarball_path`, `entrypoint`, `timeout`

Both already work today. What is missing is that the raw create path lacks three things the preset paths have, and each gap is visible to the client.

## 1. `template` provenance on `CreateAutomationRequest`

`TemplateProvenance` (`preset_router.py:122`) is accepted by both preset endpoints, stored under `preset_metadata["template"]`, and drives idempotent creation through `_find_existing_template_automation` (`preset_router.py:446`). `CreateAutomationRequest` has no equivalent, so an automation created from a catalog bundle has:

- no record of which catalog entry and version it came from,
- no idempotent create — clicking the card twice creates two automations,
- no stored config for a later edit dialog,
- no way for a client to show the entry as already enabled.

Suggested shape: move `TemplateProvenance` from `preset_router.py` to `schemas.py` (`router.py` already imports `schemas`, so this keeps the import graph acyclic), add `template: TemplateProvenance | None` to `CreateAutomationRequest`, and reuse the existing lookup for the 200-instead-of-201 path.

## 2. Preflight for the raw create body

`POST /v1/validate` resolves its model from `_DRAFT_MODELS`, keyed by endpoint (`capabilities_router.py:59`), and `ValidateDraftRequest.endpoint` is `Literal["/v1/preset/prompt", "/v1/preset/plugin"]` (`schemas.py:766`). A bundle draft therefore has no preflight path at all: a mapping mistake between the form and the request body surfaces as a 422 at creation time, which is exactly what preflight exists to prevent.

Suggested: add `"/v1"` to the `Literal` and `_DRAFT_MODELS["/v1"] = CreateAutomationRequest`. The cron/event checks in `validate_draft` are already model-agnostic — they read `draft.trigger`.

## 3. A capability flag

`_STATIC_FEATURES` (`capabilities_router.py:66`) reports `conversationDispatch`, `mcpTools`, `presetPlugin`, `presetPrompt`, `repoClone`. There is no name for "this deployment can run a user-supplied tarball", so an extensions entry cannot declare the requirement in `requires.features` and a client cannot tell whether a bundle card is runnable here.

Suggested: add `customTarball` to `_STATIC_FEATURES`.

## Notes

- Nothing here changes the execution path. A bundle run is an ordinary tarball run.
- Upload ownership is already enforced the way we need: `validate_tarball_path` requires the upload to exist, be `COMPLETED`, and belong to the same user and org as the creator (`utils/tarball_validation.py:74-103`).
- `POST /v1` already snapshots the user's active model profile when `model` is omitted (`utils/model_profiles.py:27-39`), so a bundle that starts conversations gets `AUTOMATION_MODEL` without the client passing anything.
- Note for whoever picks this up: a bundle has no `repos` field on `CreateAutomationRequest`, so it cannot ask the service to clone. That is fine for the first entry — it fetches what it needs itself — but it is worth deciding whether the raw path should gain `repos` for parity with the presets.
- Separately filed while investigating this: #343 — `setup_script_path` is validated and stored but never read by any executor.

VascoSch92:
Closes #344.

A catalog entry that ships a **script tarball** instead of a prompt creates through the ordinary raw path — `POST /v1/uploads`, then `POST /v1` — which already works. What was missing is the three things the preset paths have and the raw path did not, each visible to the client.

## 1. `template` provenance on `CreateAutomationRequest`

`TemplateProvenance` moves from `preset_router.py` to `schemas.py` (unchanged shape), and `CreateAutomationRequest` gains `template: TemplateProvenance | None`. On the raw create path it is stored under `preset_metadata["template"]`, exactly as the presets store it, which gives a bundle automation the same four things: a record of the catalog entry and version, idempotent creation (the second enable returns the existing automation with **200** instead of creating a duplicate), stored config for a later edit dialog, and a way for a client to show the entry as already enabled.

The lookup is the existing one — `_find_existing_template_automation` moves to `utils/templates.py` as `find_existing_template_automation` and is now called by all three creation endpoints, so template identity is shared across them: an entry enabled as a preset is not enabled again as a bundle. Same for `TEMPLATE_EXISTS_RESPONSE`, which was duplicated per endpoint.

The lookup runs before tarball validation, matching the presets, so a repeat enable costs one query and leaves the freshly uploaded tarball unreferenced rather than adopting it.

## 2. Preflight for the raw create body

`ValidateDraftRequest.endpoint` accepts `"/v1"` and `_DRAFT_MODELS["/v1"] = CreateAutomationRequest`. The cron/event checks in `validate_draft` were already model-agnostic and needed no change. Preflight validates the request body, not the upload behind it — `tarball_path` is checked for scheme here and for ownership at creation.

## 3. A capability flag

`customTarball` is added to `_STATIC_FEATURES`, so an extensions entry can declare the requirement in `requires.features` and a client can tell whether a bundle card is runnable on this deployment.

## Notes

- Nothing here changes the execution path. A bundle run is an ordinary tarball run.
- `record_first_run_outcome` already keys its one-time first-run telemetry on `preset_metadata["template"]`, so bundle automations get that for free.
- `automation_created` telemetry on the raw path now carries `template_id`/`template_version` when a template is present, matching the presets.
- Left out deliberately, per the issue's open question: `repos` on `CreateAutomationRequest`. `github-pr-reviewer` fetches what it needs itself. Worth deciding separately whether the raw path should gain it for parity.

## Test plan

- `tests/test_router.py::TestCreateAutomation` — provenance stored and returned; second enable returns the first automation with 200 and the listing still totals 1; a template enabled as a preset is not re-enabled as a bundle; deleting frees the entry to be enabled again; oversized `config` is a 422.
- `tests/test_capabilities_router.py` — `customTarball` advertised; a `/v1` draft preflights clean; a missing `entrypoint` is reported against its field instead of surfacing as a 422 at creation; the cron floor applies to a bundle draft unchanged.
- Full unit suite: `uv run pytest tests/ --ignore=tests/integration` → 1300 passed. Two failures in `tests/test_router.py` (`test_create_automation_success`, `test_dispatch_automation_success`) reproduce on `main` when those files are run alone and are unrelated to this change.
- `pre-commit run --files ...` (ruff format/lint, pycodestyle, pyright) passes.

### DIFF
--- a/openhands/automation/schemas.py
+++ b/openhands/automation/schemas.py
@@ -1,5 +1,6 @@
 """Pydantic request/response schemas for the API."""
 
+import json
 import re
 import uuid
 from enum import StrEnum
@@ -244,6 +245,57 @@ def validate_command_string(
     return v
 
 
+# --- Template provenance ---
+
+# Cap on the serialized size of template provenance config, so an opaque
+# payload cannot bloat the preset_metadata JSON column.
+MAX_TEMPLATE_CONFIG_BYTES = 16_384
+
+
+class TemplateProvenance(BaseModel):
+    """Opaque provenance of the extension-owned template an automation came from.
+
+    The service stores this verbatim under ``preset_metadata["template"]`` and
+    never validates it against any catalog — template definitions are owned by
+    OpenHands/extensions. Must not contain secrets.
+
+    Lives here rather than in ``preset_router`` because every creation path
+    accepts it: the two preset endpoints and the raw ``POST /v1`` a catalog
+    entry shipping its own tarball uses.
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    id: str = Field(
+        ...,
+        min_length=1,
+        max_length=100,
+        description="Identifier of the extension template (catalog entry id).",
+    )
+    version: str = Field(
+        ...,
+        min_length=1,
+        max_length=50,
+        description="Version of the template at creation time.",
+    )
+    config: dict[str, Any] | None = Field(
+        default=None,
+        description=(
+            "Non-secret configuration the user submitted when enabling the "
+            "template (e.g. setup form values)."
+        ),
+    )
+
+    @field_validator("config")
+    @classmethod
+    def validate_config_size(cls, v: dict[str, Any] | None) -> dict[str, Any] | None:
+        if v is not None and len(json.dumps(v)) > MAX_TEMPLATE_CONFIG_BYTES:
+            raise ValueError(
+                f"config must serialize to at most {MAX_TEMPLATE_CONFIG_BYTES} bytes"
+            )
+        return v
+
+
 # --- Requests ---
 
 
@@ -285,6 +337,15 @@ class CreateAutomationRequest(BaseModel):
             "If true, leave the sandbox for runtime TTL cleanup after the run "
             "finishes. If false or null, explicitly clean it up after "
             "completion (or after post-run callbacks, when configured)."
+        ),
+    )
+    template: TemplateProvenance | None = Field(
+        default=None,
+        description=(
+            "Opaque provenance of the extension template this automation is "
+            "created from. Enables idempotent creation: when a live automation "
+            "for the same user and template id already exists, it is returned "
+            "unchanged with HTTP 200 instead of creating a duplicate."
         ),
     )
 
@@ -764,8 +825,12 @@ class ValidateDraftRequest(_SetupContractModel):
 
     model_config = ConfigDict(extra="forbid")
 
-    endpoint: Literal["/v1/preset/prompt", "/v1/preset/plugin"] = Field(
-        ..., description="Creation endpoint the draft will be sent to"
+    endpoint: Literal["/v1", "/v1/preset/prompt", "/v1/preset/plugin"] = Field(
+        ...,
+        description=(
+            "Creation endpoint the draft will be sent to. '/v1' is the raw "
+            "create path a catalog entry shipping its own tarball uses."
+        ),
     )
     draft: dict[str, Any] = Field(
         ..., description="The request body that would be sent to that endpoint"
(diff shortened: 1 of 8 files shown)


openhands/automation/schemas.py
```diff
@@ -244,6 +245,57 @@ def validate_command_string(
     return v
 
 
+# --- Template provenance ---
+
+# Cap on the serialized size of template provenance config, so an opaque
+# payload cannot bloat the preset_metadata JSON column.
+MAX_TEMPLATE_CONFIG_BYTES = 16_384
```
VascoSch92: can you use Final[int] here thanks

VascoSch92: Done in c616b9e — `MAX_TEMPLATE_CONFIG_BYTES: Final[int] = 16_384`, matching `git_sync/secret_store.py` and `utils/periodic_loop.py`.

Trimmed the prose across the PR in the same commit: the `TemplateProvenance` docstring, the two field descriptions, the preflight and `customTarball` comments, the idempotency comment in `create_automation`, the `utils/templates.py` module docstring and lookup docstring, and the AGENTS.md section.

neubig (approved):
Basically LGTM, thank you!

Issue #363, opened 2026-08-21
Tracking: one ingestion pipeline for webhook and Socket Mode event transports


VascoSch92:
## Background

Event-triggered automations work today. `POST /v1/events/{org_id}/{source}`
(`event_router.py:118–314`) reads the raw body, looks up the webhook config,
verifies an HMAC signature, parses the provider payload into a typed
`WebhookEvent`, derives an `event_key`, finds the org's event automations,
runs `matches_trigger()` against each, creates `PENDING` `AutomationRun` rows,
and commits. The dispatcher picks them up from there.

Two parts of that are already transport-agnostic and good:

- `trigger_matcher.matches_trigger(trigger, source, event_key, payload)` —
  source equality, event-key wildcards, JMESPath filter. Knows nothing about HTTP.
- `EventTrigger` (`type/source/on/filter`) — the external contract an automation
  author writes. It never mentions how the event arrived, which is correct.

Everything below `matches_trigger()` is reusable as-is. The problem is above it.

## Problem

`receive_event()` fuses three separate concerns into one HTTP handler:

1. **Acquisition** — how the bytes arrived
2. **Authentication** — proving the event is genuine
3. **Interpretation** — payload → `event_key` (+ subject)

and hard-wires (2) to GitHub-style HMAC-SHA256 (`utils/webhook.py:67`). A second
transport therefore has nowhere to plug in, and would have to duplicate trigger
matching and run creation to get anything done — which is exactly what draft
PR #134 does inside `SocketManager`.

## Why now

We already run the target shape in production, outside the repo. The OSS
automation VM has run Slack Socket Mode since 2026-08-13 as a separate systemd
unit (`slack-socket-bot.service`, ~267 lines, `NRestarts=0`, 8 days continuous).
It holds one Socket Mode connection per Slack app, acks, filters, then HMAC-signs
the raw Slack envelope and POSTs it to the *existing* `/v1/events/{org_id}/slack`.

Two automations consume it with ordinary triggers:

```json
{"type": "event", "source": "slack", "on": "app_mention", "filter": "team_id == 'T06P212QSEA'"}
```

So socket events already traverse the identical verify → parse → match →
create-run path as GitHub webhooks, with zero special-casing in the service.
The shared pipeline is proven end-to-end. What remains is to move that boundary
from an HTTP hop into a function call, and to generalise the three fused seams.

## Target

| Seam | Today | Target |
|---|---|---|
[...]

Issue #362, opened 2026-08-21
Route events to an existing conversation via an external subject mapping


VascoSch92:
> **Phase 5 of 5** · part of #363
>
> **Blocked by:** #358 and #359 — both closed, so this is unblocked.
> **Blocks:** nothing. #358–#361 do not depend on this.
>
> One nullable column on `automation_runs`. No new table.

## Background

Every matched event creates a brand-new `AutomationRun`, and therefore a brand-new
conversation with no memory of what came before.

On the OSS VM this is visible today: each `app_mention` in a Slack thread starts a
fresh run. Someone can @-mention the bot, get an answer, reply in-thread with a
follow-up, and the bot answers the follow-up as though it had never seen the
thread. The same shape applies to repeated events on one GitHub PR or one Linear
issue.

PR #118 attempted this with a `SessionConfig` on `EventTrigger`, JMESPath
session-key extraction, an `AutomationSession` model, a `PendingSessionEvent`
queue, and active/dead/expired session states. Its diagnosis of the user problem
is right. Its mechanism makes the automation sandbox a long-lived event consumer,
and the PR itself notes it still needs agent-server event queue endpoints, an SDK
polling loop inside the sandbox, and sandbox restart with event requeue.

That is a substantial amount of lifecycle machinery to add inside a system whose
central abstraction is a finite run.

## Problem

Threaded, multi-turn interactions over an external subject (a Slack thread, a PR,
an issue) cannot share context, and the obvious fix builds a second conversation
system inside the sandbox.

## Proposed change

Keep conversations where they already live, and **derive** the conversation id
instead of storing a mapping.

**Principle: the id is a documented, deterministic function of the external
subject.** Nothing is looked up to compute it, so there is no table, no
migration, and nothing that can dangle or need expiring.

```
conversation_id = uuid5(
    AUTOMATION_CONVERSATION_NS,
    f"{org_id}/{automation_id}/{source}/{subject_key}",
)
```

`AUTOMATION_CONVERSATION_NS` is pinned permanently — changing it orphans every
live thread. `automation_id` is in the key on purpose: editing an automation
re-keys its threads. Without it a thread stays pinned to whatever agent and
config the first event saw, and attaching with a different agent kind raises
(`_agent_kind_mismatch_message`).

Providers emit an optional `subject` on `AcceptedEvent` (the field was reserved in
phase 1, and the extractor hook in phase 2's descriptor):

- Slack — `team/channel/thread_ts`
[...]

PR #367, opened 2026-08-23
refactor: extract a transport-neutral accept_event() from the webhook handler

### LINKED ISSUE #358: Extract a transport-neutral accept_event() from the webhook handler
> **Phase 1 of 5** · part of #363
>
> **Blocked by:** nothing. This can start immediately.
> **Blocks:** #360, #361, #362 — all three need this seam to exist.
>
> Pure refactor: no behaviour change, no schema change, no new dependencies.

## Background

`receive_event()` (`event_router.py:118–314`) is currently one function doing
seven things in sequence:

```
137  body = await request.body()                 # raw body, for HMAC
141  config = await get_webhook_config(...)      # builtin secret or per-org CustomWebhook
155  signature = request.headers.get(config.signature_header)
169  verify_signature(body, signature, config.secret)
179  payload = await request.json()
187  event = parse_event(source, payload, ...)   # typed WebhookEvent + event_key
---- ------------------------------------------- boundary
247  automations = await get_event_automations(org_id, source, session)
252  matches_trigger(trigger, source, event.event_key, webhook_payload)
284  run = await create_automation_run(automation, session, event_payload=...)
308  await session.commit()
```

Lines 137–244 are HTTP-specific: they exist to turn an untrusted HTTP request
into a trustworthy, interpreted event. Lines 246–313 are not HTTP-specific at
all — they are the routing and run-creation logic every transport needs.

## Problem

Because the two halves live in one function, a second transport has nowhere to
attach. It must either re-implement matching and run creation (what draft PR #134
does inside `SocketManager`) or fake an HTTP request to reach them (what the OSS
VM's Slack bridge does — it HMAC-signs a body and POSTs to loopback purely to get
past the handler's front half).

Both are workarounds for a missing function boundary.

## Proposed change

Add `openhands/automation/ingest.py`:

```python
@dataclass
class AcceptedEvent:
    source: str
    event_key: str
    payload: dict                       # raw provider payload; JMESPath filters run on this
    provider_event_id: str | None = None
    subject: EventSubject | None = None
    occurred_at: datetime | None = None

@dataclass
class AcceptResult:
    matched: int
    run_ids: list[str]
    duplicate: bool = False             # unused until phase 4

async def accept_event(
    org_id: uuid.UUID,
    event: AcceptedEvent,
    session: AsyncSession,
    *,
    request: Request | None = None,     # telemetry only
) -> AcceptResult: ...
```

Move `event_router.py:246–313` into it verbatim. `receive_event()` keeps lines
137–244 and ends with a call to `accept_event()`, mapping `AcceptResult` onto the
existing `EventResponse`.

### Why a new module rather than `utils/webhook.py`

`utils/webhook.py` is HTTP-flavoured — it holds `verify_signature()`,
`get_webhook_config()`, and the `BUILTIN_SOURCES` secret registry. Putting the
deliberately transport-neutral boundary inside a file named `webhook.py` works
against the point of the change, and the next person adding a transport will read
the filename and conclude the function is not for them. `ingest.py` can import
`get_event_automations` and `create_automation_run` from `utils/webhook.py` for
now; relocating those later is cosmetic and can happen in any follow-up.

### Notes for the implementer

- `capture_automation_event` already accepts `request: Request | None = None`
  (`telemetry.py:391`), so non-HTTP callers simply omit it and all existing
  telemetry events keep firing. No telemetry changes are needed.
- Keep the existing `event_payload` behaviour exactly: `event.model_dump(mode="json")`
  for Pydantic-parsed events (GitHub), raw `webhook_payload` for custom webhooks
  (`event_router.py:276–280`). This is subtle and easy to get wrong.
- `AcceptedEvent.subject` and `provider_event_id` are unused in this phase. Add
  the fields now so later phases do not have to change the signature, but do not
  populate or act on them yet.

## Acceptance criteria

- [ ] The existing `event_router` test suite passes **unmodified.** This is the
      entire proof of behaviour preservation and should not be negotiated away by
      "just updating the tests".
- [ ] `EventResponse` is byte-identical on the wire for all existing sources.
- [ ] No schema change, no new dependency, no change to `trigger_matcher.py`,
      `create_automation_run()`, or the dispatcher.
- [ ] `accept_event()` has direct unit tests that construct an `AcceptedEvent`
      without going through HTTP.

## Out of scope

Deduplication, event persistence, and any new transport. This phase only creates
the seam; phases 3 and 4 use it.

VascoSch92:
Closes #358 — phase 1 of 5, part of #363.

Pure refactor: no behaviour change, no schema change, no new dependencies.

## What changed

`receive_event()` was one function doing two unrelated jobs. Lines 137–244 are
HTTP-specific — they turn an untrusted request into a trustworthy, interpreted
event. Lines 246–313 are not HTTP-specific at all — they are the routing and
run-creation logic every transport needs. Because both lived in one function, a
second transport had nowhere to attach: it had to re-implement matching and run
creation (draft PR #134), or fake an HTTP request to reach them (what the OSS
VM's Slack bridge does — HMAC-signing a body and POSTing to loopback purely to
get past the handler's front half).

This adds `openhands/automation/ingest.py` with the boundary:

```python
async def accept_event(
    org_id: uuid.UUID,
    event: AcceptedEvent,
    session: AsyncSession,
    *,
    request: Request | None = None,   # telemetry only
) -> AcceptResult
```

The routing half moves there unchanged. `receive_event()` keeps acquisition,
authentication and interpretation, then calls `accept_event()` and maps
`AcceptResult` onto the existing `EventResponse`. Net effect on the router:
−65 lines, one call.

Authentication is deliberately not in this module, per the design rule in #363:
`accept_event()` receives an already-authenticated event and never asks how it
was authenticated.

## ⚠️ One correction to the issue: `accept_event()` needs a `session_factory`

Issue #358 says:

> `capture_automation_event` already accepts `request: Request | None = None`
> (`telemetry.py:391`), so non-HTTP callers simply omit it and all existing
> telemetry events keep firing. No telemetry changes are needed.

**That is not correct, and phase 3 would have shipped with no telemetry.**
Telemetry resolves a backend distinct id from the database. With no `request`,
no `session` and no `session_factory`, `get_automation_backend_distinct_id()`
returns `None` (`telemetry.py:75–77`), and `_capture_automation_event` then
returns before the PostHog POST. `capture_automation_event` swallows every
exception (`telemetry.py:395–396`), so the loss is completely silent.

Measured, calling `capture_automation_event` exactly as the seam does:

```
distinct_id with no request/session = None
distinct_id with session=           = 'automation-backend:0b72aac9-…'
posts after request=None call: []          <- all events dropped
posts after session= call:     ['automation_event_matched']
```

A transport calling `accept_event(..., request=None)` loses all three events per
run (`automation_event_matched`, `automation_run_scheduled`,
`automation_run_created`) while routing keeps working — which is exactly what
makes it easy to miss.

So `accept_event()` takes an optional `session_factory` and forwards it. The
HTTP path passes nothing and reaches the factory via
`request.app.state.session_factory` as before, so it is byte-identical.

**Why not just pass `session`?** It would work, but
`_get_or_create_backend_distinct_id` issues a raw `INSERT … ON CONFLICT DO
NOTHING` on whatever session it is handed (`telemetry.py:40–58`). Today that
runs on a *fresh* session committed immediately (`telemetry.py:79–82`). Sharing
the caller's session moves that write into the caller's transaction, and a
failure there would be hidden by the same broad `except` while poisoning the
`create_automation_run()` and `commit()` that follow.

This also means the issue's `# telemetry only` comment on `request` understates
it: `request` is load-bearing for a DB lookup, not just metadata.

## Two implementation notes worth reviewing

**1. `parsed_event` — the field the issue didn't specify, and why it's needed.**

The issue flags that `event_payload` behaviour is subtle and easy to get wrong:
`event.model_dump(mode="json")` for Pydantic-parsed events, raw
`webhook_payload` otherwise (`event_router.py:276–280`). But the proposed
`AcceptedEvent` carries only `payload` — the raw provider payload that JMESPath
filters run against — which for GitHub is *not* what gets persisted on the run.
There is no way to reproduce the existing behaviour from the six specified
fields alone.

So `AcceptedEvent` gains one additional optional field, `parsed_event`, holding
the typed `WebhookEvent` the HTTP handler already builds. All six fields from
the issue are present, in order, with the specified defaults; this is additive
and defaulted, so it doesn't change the signature for later phases. Transports
with no typed model simply omit it and their raw `payload` is persisted, which
is the documented fallback behaviour.

Worth noting for anyone checking the branch: `parse_event()` always returns a
`WebhookEvent`, which is a `BaseModel`, so the `else webhook_payload` branch is
in practice unreachable from the HTTP path today. The `isinstance` check is
preserved verbatim rather than simplified — this is a refactor, and collapsing
it would be a behaviour decision that belongs in its own change.

**2. `EventSubject` did not exist.**

It is referenced by the issue's dataclass but is defined nowhere in the repo,
and #362 (which consumes it) only describes it as producing a `subject_key`.
It is defined here minimally — a single `key: str` — so the reserved field has
a real type. Nothing populates or reads it.

## Acceptance criteria

- [x] **The existing `event_router` test suite passes unmodified.** 14/14, with
      zero changes to `tests/test_event_router.py`.
- [x] **`EventResponse` is byte-identical on the wire.** Verified by generating
      the full OpenAPI spec on `main` and on this branch and diffing them —
      identical, 160,253 bytes each.
- [x] **No schema change, no new dependency, no change to `trigger_matcher.py`,
      `create_automation_run()`, or the dispatcher.** The diff is three files:
      one new module, one new test file, and the router.
- [x] **`accept_event()` has direct unit tests that construct an `AcceptedEvent`
      without going through HTTP.** 16 tests in `tests/test_ingest.py` — no
      request, no signature, no FastAPI app. They cover no-match, match,
      per-automation run fan-out, filter evaluation against `payload`, event-key
      mismatch, source and org isolation, commit visibility from a second
      session, both `event_payload` shapes, the full telemetry sequence for a
      request-less caller, that the reserved fields change nothing, and that the
      dataclasses are frozen and slotted.

All three dataclasses are `frozen=True, slots=True`. Two caveats, since frozen
is easy to over-read: it is shallow, so `payload` and `run_ids` are still
mutable in place; and it generates a `__hash__` that raises at runtime for
`AcceptedEvent`/`AcceptResult` because of those dict/list fields. Phase 4's
dedupe should key on `provider_event_id`, not on the event object.

Full suite: 1350 passed. `pre-commit` (ruff format, ruff lint, pycodestyle,
pyright) clean on all three files.

## One question for the author

The `else webhook_payload` branch is already dead. `parse_event()` returns
`WebhookEvent` (`event_schemas/__init__.py:22`, a `BaseModel`) and
`CustomWebhookEvent` subclasses it (`custom.py:72`), so `isinstance(event,
BaseModel)` is always true. Custom webhooks today persist `{"payload": …,
"source_override": …, "event_key": …}`, **not** the raw payload that the code
comment and the issue both describe. I preserved the branch verbatim, since
collapsing it is a behaviour decision — but phases 4 and 5 should not assume the
documented behaviour is what actually happens.

## Out of scope

Deduplication, event persistence, and any new transport — phases 3 and 4 use
this seam.

### DIFF
--- a/openhands/automation/ingest.py
+++ b/openhands/automation/ingest.py
@@ -1,22 +1,8 @@
-"""
-Transport-neutral event ingestion.
+"""Transport-neutral event ingestion.
 
-`accept_event()` is the boundary between *how an event arrived* and *what the
-service does with it*. Everything above the boundary — reading bytes, proving
-the event is genuine, turning a provider payload into an ``event_key`` — belongs
-to the transport. Everything below it — finding the org's event automations,
-matching triggers, creating PENDING runs — is identical for every transport and
-lives here.
-
-Authentication is deliberately **not** part of this module. HTTP proves
-authenticity with an HMAC over the raw body; a Socket Mode connection proves it
-by possession of an app-level token over TLS; a poller proves it by having made
-the outbound call itself. `accept_event()` receives an already-authenticated
-event and never asks how it was authenticated. Without that rule, every
-non-HTTP transport is tempted to manufacture a signature to satisfy a check that
-does not apply to it.
-
-The only caller today is the webhook handler in `event_router.py`.
+`accept_event()` separates how an event arrived from what the service does with
+it. Transports own acquisition, authentication and interpretation; trigger
+matching and run creation live here and are shared by every transport.
 """
 
 import logging
@@ -40,40 +26,29 @@ from openhands.automation.utils.webhook import (
 logger = logging.getLogger("automation.ingest")
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
 class EventSubject:
-    """The external thing an event is about (a Slack thread, a pull request).
-
-    Reserved for subject -> conversation routing; nothing populates or reads it
-    yet. `key` is the value that will be stored as `subject_key`.
-    """
+    """The external thing an event is about. Reserved; nothing reads it yet."""
 
     key: str
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
 class AcceptedEvent:
-    """An authenticated, interpreted event, ready to be routed.
-
-    The transport owns everything that produces one of these: acquiring the
-    bytes, proving they are genuine, and deriving `event_key`.
-    """
+    """An authenticated, interpreted event, ready to be routed."""
 
     source: str
     event_key: str
-    # Raw provider payload. JMESPath trigger filters run against this.
+    # Raw provider payload; JMESPath trigger filters run on this.
     payload: dict[str, Any] = field(default_factory=dict)
     provider_event_id: str | None = None
     subject: EventSubject | None = None
     occurred_at: datetime | None = None
-    # Optional typed event, when the transport parsed the payload into a
-    # Pydantic model (the webhook handler does this via `parse_event()`). When
-    # present it — not `payload` — is what gets persisted as the run's
-    # `event_payload`, preserving the shape existing automations already see.
+    # When set, persisted as the run's event_payload in place of `payload`.
     parsed_event: BaseModel | None = None
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
 class AcceptResult:
     """The outcome of routing an accepted event."""
 
@@ -89,29 +64,18 @@ async def accept_event(
     *,
     request: Request | None = None,
 ) -> AcceptResult:
-    """
-    Route an already-authenticated event to the org's matching automations.
+    """Route an already-authenticated event to the org's matching automations.
 
-    Args:
-        org_id: The organization the event belongs to
-        event: The authenticated, interpreted event
-        session: Database session (committed before returning)
-        request: Originating HTTP request, for telemetry only. Non-HTTP
-                 transports omit it; every telemetry event still fires.
-
-    Returns:
-        AcceptResult with the number of matched automations and the IDs of the
-        runs created for them.
+    Commits before returning. `request` is telemetry only; non-HTTP transports
+    omit it and every telemetry event still fires.
     """
     source = event.source
     webhook_payload = event.payload
 
-    # 1. Find matching automations
     automations = await get_event_automations(org_id, source, session)
     matched_automations = []
 
     for automation, trigger in automations:
-        # Match trigger against webhook payload using JMESPath filter
         if matches_trigger(trigger, source, event.event_key, webhook_payload):
             matched_automations.append(automation)
 
@@ -133,9 +97,7 @@ async def accept_event(
         },
     )
 
-    # 2. Create PENDING runs for matched automations
-    # For Pydantic-parsed events (GitHub), use model_dump() for typed fields
-    # For custom webhooks, use the webhook payload directly
+    # Typed events (GitHub) keep their model shape; others store the payload.
     event_payload = (
         event.parsed_event.model_dump(mode="json")
         if isinstance(event.parsed_event, BaseModel)
--- a/openhands/automation/ingest.py
+++ b/openhands/automation/ingest.py
@@ -13,7 +13,7 @@ from typing import Any
 
 from fastapi import Request
 from pydantic import BaseModel
-from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
 
 from openhands.automation.telemetry import capture_automation_event
 from openhands.automation.trigger_matcher import matches_trigger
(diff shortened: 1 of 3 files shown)


malhotra5 (approved):
LGTM

openhands/automation/ingest.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+logger = logging.getLogger("automation.ingest")
+
+
+@dataclass(frozen=True, slots=True)
+class EventSubject:
+    """The external thing an event is about. Reserved; nothing reads it yet."""
+
+    key: str
```
malhotra5, line 33: What is this meant for?

VascoSch92: It is justa. placeholder that I will need later.

Issue #360, opened 2026-08-21
Add in-service Slack Socket Mode via a supervised stream transport


VascoSch92:
> **Phase 3 of 5** · part of #363
>
> **Blocked by:** #358 — hard. The provider calls `accept_event()`; without it there is nothing to emit into.
> **Also needs:** #359 — soft. Can land single-replica without the capability flag, but not multi-replica.
> **See also:** #361 — stream sources are only safe on a single replica until persistent dedupe exists.
> **Blocks:** nothing.
>
> No schema change, no migration.

## Background

Slack Socket Mode already works in production — but outside this repository.

The OSS automation VM runs `slack-socket-bot.service`, a ~267-line Python script
(`/opt/slack-socket-bot/slack_socket_bridge.py`) in its own venv, as its own
systemd unit with `Restart=always` and `Requires=agent-canvas.service`. It:

1. holds one `slack_sdk` Socket Mode connection per Slack app (currently two:
   `openhands-ai` / `T06P212QSEA`, and `all-hands-ai` / `T07E1S2M2Q6`)
2. acks the envelope immediately, before doing any other work
3. filters to `app_mention`, drops bot messages, deduplicates on `event_id`,
   and asserts the envelope's `team_id`
4. HMAC-signs the raw Slack envelope and POSTs it to the **existing**
   `/v1/events/{org_id}/slack` endpoint with `X-Signature-256`

The automation service treats it as an ordinary custom webhook. Two automations
consume it with completely ordinary triggers:

```json
{"type": "event", "source": "slack", "on": "app_mention", "filter": "team_id == 'T06P212QSEA'"}
```

Operational record: running since 2026-08-13, `NRestarts=0`, 8 days continuous,
zero unmatched events, zero duplicate deliveries observed.

This is strong evidence that the design is right — socket events need no
special-casing anywhere in the pipeline. What is wrong is only *where the code
lives*.

## Problem

**1. The loopback hop silently drops events.** The bridge acks Slack first (it
must — Slack redelivers un-acked envelopes) and then forwards on a daemon thread
with no retry. Two forwards have failed in eight days:

```
Aug 17 14:23:42  [openhands-ai] forward FAILED: timed out
Aug 17 14:41:41  [openhands-ai] forward FAILED: timed out
```

Those are two people who @-mentioned the bot and got nothing. Slack will not
redeliver, because the envelope was already acked. In-process there is no POST,
no 30-second timeout and no thread — **this failure mode ceases to exist**, which
[...]

PR #378, opened 2026-08-23
feat: replace the parse-only source registry with a provider descriptor and verifier registry

### LINKED ISSUE #359: Replace the parse-only source registry with a provider descriptor and verifier registry
> **Phase 2 of 5** · part of #363
>
> **Blocked by:** nothing. Independent of #358 — the two can run in parallel.
> **Blocks:** #360 (needs `capabilities.tolerates_multiple_connections`), #362 (needs the provider `subject` extractor hook).
>
> One nullable column; no behaviour change for existing sources.

## Background

The service describes an event source in two unrelated places:

- **`BUILTIN_SOURCES`** (`utils/webhook.py:39–44`) — maps a source slug to a
  function extracting its shared secret from `Settings`. Currently `github`,
  `jira_dc`, `bitbucket_data_center`. Also gates `RESERVED_SOURCES`, so custom
  webhooks cannot impersonate a built-in provider.
- **`event_schemas`** registry — maps a source slug to a parser producing a typed
  `WebhookEvent` with an `event_key`. Unknown sources fall back to
  `CustomWebhookEvent` with a configurable JMESPath expression.

These are two halves of one concept — "what do we know about this provider" —
split across two files, so adding a provider means editing both and remembering
`RESERVED_SOURCES` as well.

Separately, signature verification is not actually general. `verify_signature()`
(`utils/webhook.py:67`) is hard-coded to HMAC-SHA256 over the raw body, accepting
either `sha256=<hex>` or bare hex. `CustomWebhook` lets an operator configure the
*header name* (`signature_header`) but nothing else: not the algorithm, not the
encoding, not a timestamp header or replay window, not the signed-content format,
not multiple signatures. Anything that does not sign the raw body with
hex-encoded HMAC-SHA256 simply cannot be onboarded.

PR #247 (closed, unmerged) diagnosed this correctly and started on the fix. Its
core idea — a signature *scheme* resolved per source rather than a single
hard-coded algorithm — is what this issue revives.

## Problem

1. Adding a provider touches two registries plus a reserved-name list.
2. Providers whose signing differs from GitHub's cannot be supported at all.
   Slack (`v0:{timestamp}:{body}`), Stripe, and anything implementing
   Standard Webhooks are all currently out of reach.
3. There is no hook for providers that require a handshake before the first real
   event (a challenge/validation response).
4. Nothing declares provider *capabilities*, which phase 3 needs — specifically
   whether a provider tolerates multiple simultaneous connections.

## Proposed change

One descriptor in a new `openhands/automation/providers.py`:

```python
class WebhookVerifier(Protocol):
    def verify(self, *, body: bytes, headers: Mapping[str, str], secret: str) -> bool: ...

VERIFIERS: dict[str, WebhookVerifier] = {
    "hmac_sha256_hex":   ...,  # current verify_signature(), incl. bare-hex (Linear)
    "standard_webhooks": ...,  # from PR #247 — covers more than one vendor
    "slack_v0":          ...,  # v0:{ts}:{body}, with timestamp freshness
}

@dataclass(frozen=True)
class Provider:
    source: str
    parse: Callable[..., WebhookEvent]
    verifier: str = "hmac_sha256_hex"
    secret_from_settings: Callable[[Settings], str | None] | None = None
    subject: Callable[[dict], EventSubject | None] | None = None   # unused until phase 5
    handshake: Callable[[Request], Response | None] | None = None
    capabilities: Capabilities = Capabilities()
```

`Capabilities` starts with a single field that phase 3 depends on:
`tolerates_multiple_connections`. Slack load-balances across up to 10 Socket Mode
connections and is therefore safe to run from multiple replicas; an arbitrary
WebSocket upstream would broadcast to all of them and duplicate every event.
That difference must be declared, not assumed.

`event_router.py` resolves `VERIFIERS[provider.verifier]` instead of calling
`verify_signature()` directly. `RESERVED_SOURCES` derives from the descriptor
registry rather than being maintained separately.

### Schema

`custom_webhooks` gains a nullable `signature_scheme` column defaulting to
`hmac_sha256_hex`. Existing rows are unaffected.

### Backward compatibility

Wrap the three existing parsers in descriptors using the default verifier and
their current `secret_from_settings` lambdas. `github`, `jira_dc` and
`bitbucket_data_center` must behave identically — same secrets, same headers,
same parse results, same `event_key`s.

### On the handshake hook

The architecture discussion used Slack's Events API challenge as the motivating
example for handshakes. If Socket Mode becomes our Slack path (phase 3), that
example no longer applies. Add the hook only if another provider on the roadmap
actually needs it; otherwise defer it and keep the descriptor field reserved.
Worth deciding explicitly rather than building it on autopilot.

## Acceptance criteria

- [ ] `github`, `jira_dc`, `bitbucket_data_center` behave identically —
      existing tests pass unmodified.
- [ ] Existing `CustomWebhook` rows verify exactly as before (scheme defaults to
      `hmac_sha256_hex`).
- [ ] A custom webhook can be configured with `standard_webhooks` and verifies a
      known-good fixture, including timestamp rejection outside the replay window.
- [ ] Adding GitLab (PR #66) requires one new file and a registry entry — no
      changes to `event_router.py`.
- [ ] `RESERVED_SOURCES` is derived, not hand-maintained.

## Out of scope

Stream transports (phase 3) and any use of `subject` (phase 5). This phase only
widens the description of a provider.

## References

- PR #247 — signature schemes (closed; correct diagnosis, revive the verification work)
- PR #66 — GitLab built-in events (the test case for "is this now easy?")

VascoSch92:
Closes #359 — phase 2 of 5, part of #363.

Phase 1 (#367) is merged, so this is now rebased onto `main` and the diff is
phase 2 only. Independent of phase 1 in substance — the two issues had no
blocking relationship — but they touch adjacent code, so it was developed
stacked.

One nullable column. No behaviour change for any existing source or any
existing row.

## What changed

A provider was described in three places: `BUILTIN_SOURCES` (`utils/webhook.py`)
mapped a slug to a secret extractor, the `event_schemas` `_PARSERS` registry
mapped the same slug to a parser, and `RESERVED_SOURCES` (`schemas.py`) listed
the same three slugs by hand so a custom webhook could not impersonate one.
Adding a provider meant editing all three and remembering the third.

Verification was not a registry at all. `verify_signature()` was hard-wired to
hex HMAC-SHA256 over the raw body; `CustomWebhook` let an operator configure the
*header name* and nothing else. Anything signing something other than the raw
body — Standard Webhooks, Slack — could not be onboarded.

New `openhands/automation/providers.py` holds both registries:

```python
VERIFIERS: dict[str, WebhookVerifier] = {
    "hmac_sha256_hex":   HmacSha256HexVerifier(),    # unchanged behaviour, incl. bare hex
    "standard_webhooks": StandardWebhooksVerifier(), # from PR #247
    "slack_v0":          SlackV0Verifier(),          # v0:{ts}:{body}, timestamp checked
}

@dataclass(frozen=True, slots=True)
class Provider:
    source: str
    parse: ParseFunc
    verifier: str = DEFAULT_VERIFIER
    signature_header: str = DEFAULT_BUILTIN_SIGNATURE_HEADER
    secret_from_settings: SecretFunc | None = None
    subject: SubjectFunc | None = None          # reserved for #362
    handshake: HandshakeFunc | None = None      # reserved; see below
    capabilities: Capabilities = Capabilities()
```

`event_router.py` resolves `VERIFIERS[config.signature_scheme]` instead of
calling `verify_signature()`. `parse_event()` reads the descriptor, so the
parser registry is gone. `RESERVED_SOURCES` is derived. `BUILTIN_SOURCES`,
`register_builtin_source()` and `register_parser()` are removed; `verify_signature`
and `is_builtin_source` are re-exported from `utils/webhook.py` so existing
imports keep working.

`custom_webhooks.signature_scheme` is added nullable (migration `018`), and a
`NULL` is read as `hmac_sha256_hex` — the behaviour the row was created with.

## Four decisions worth reviewing

**1. The verifier protocol needed a fifth parameter.**

The issue specifies:

```python
def verify(self, *, body: bytes, headers: Mapping[str, str], secret: str) -> bool: ...
```

That cannot express the scheme it is meant to replace. `hmac_sha256_hex` reads
its signature from a header whose *name is per-row* (`signature_header`), so a
verifier given only `headers` has nothing to look the signature up by. The
implemented signature adds `signature_header: str`.

The split is deliberate and I think it is the right line: the header carrying
the **signature** is operator-configurable, so it is a parameter; the headers a
scheme needs *besides* the signature (`webhook-id`, `webhook-timestamp`,
`X-Slack-Request-Timestamp`) are fixed by the scheme, so verifiers read those
from `headers` themselves and no caller has to know about them.

**2. `slack` is deliberately NOT registered as a provider — only `slack_v0` is
added as a verifier.**

This one is a live-traffic hazard, so flagging it explicitly. The OSS automation
VM has been posting to `/v1/events/{org_id}/slack` since 2026-08-13 (#363, "Why
now"), and `slack` there is an ordinary **`CustomWebhook` row** with a per-org
secret. Registering a `slack` provider would make `get_webhook_config()` treat
that source as built-in, read `AUTOMATION_WEBHOOK_SECRET` instead of the row's
secret, and reserve the name — breaking the bridge and both automations
consuming it on the next deploy. Slack becomes a provider in #360, where the
cutover is designed. Here it only gains a verifier a custom webhook can select.

**3. The handshake hook is deferred, and the field is reserved.**

The issue asks for this to be decided rather than built on autopilot. Decision:
**defer.** The motivating example was Slack's Events API challenge, and #360
makes Socket Mode our Slack path, where it does not apply. Nothing else on the
roadmap — GitLab (#66), Stripe, Standard Webhooks, the two existing forwarded
sources — performs an HTTP handshake. The field stays on the descriptor, typed
under `TYPE_CHECKING` so `providers.py` keeps no runtime FastAPI import, with
the reasoning in a comment next to it. A test asserts no registered provider
sets it, so the deferral stays true rather than quietly decaying.

**4. `Provider.signature_header` is a field the issue did not list.**

`get_webhook_config()` hard-coded `"X-Hub-Signature-256"` for every built-in
source, with the comment `# GitHub's header`. Deriving that from the descriptor
is the difference between "adding a provider is a registry entry" and "adding a
provider is a registry entry unless it signs into a different header". Default
is the existing constant, so all three built-ins are byte-identical.

## An unrelated pre-existing test bug this surfaced

`tests/test_router.py::test_create_automation_success` and
`::test_dispatch_automation_success` assert on a browser-supplied telemetry
distinct id, which `telemetry._trusted_telemetry_context()` discards outside
local mode. They never set local mode. They pass today only because
`tests/test_local_mode.py` runs earlier and leaves a local-mode `Settings` in
the `lru_cache` — `monkeypatch` reverts the env var but not the cached object.

Confirmed pre-existing, not caused by this branch: on `vasco/accept-event-seam`
with no changes, `pytest tests/test_router.py` alone fails both. `pytest
tests/test_local_mode.py tests/test_router.py` passes.

`tests/test_providers.py` sorts between `test_preset_router.py` and
`test_router.py` and clears the config cache around each of its tests, which
removes the leak and turns the latent failure into a real one. Fixed at the
source with a `local_mode` fixture on those two tests, so they now pass standalone
too. This is the only change here to a test file unrelated to the issue; I would
rather fix the dependency than leave the suite order-fragile.

## Acceptance criteria

- [x] **`github`, `jira_dc`, `bitbucket_data_center` behave identically —
      existing tests pass unmodified.** `test_event_router.py`,
      `test_event_schemas.py`, `test_webhook_utils.py` and
      `test_webhook_router.py` are untouched and green, including
      `test_webhook_router.py:215`'s `RESERVED_SOURCES == {...}` assertion
      against the now-derived value.
- [x] **Existing `CustomWebhook` rows verify exactly as before.** Tested three
      ways: `get_webhook_config()` on a `NULL` column, a full HTTP delivery
      against a `NULL`-column row, and the migration itself — a row inserted at
      revision `016`, before the column existed, reads `hmac_sha256_hex` after
      `upgrade head` on real Postgres.
- [x] **A custom webhook can be configured with `standard_webhooks` and verifies
      a known-good fixture, including timestamp rejection outside the replay
      window.** End-to-end through `POST /v1/events/{org}/{source}`: a valid
      delivery matches an automation; a correctly signed one stamped 10,000s ago
      is refused 401. Unit tests additionally cover that the timestamp and the
      message id are *signed* (re-stamping a captured delivery fails), key
      derivation from `whsec_`, multi-signature rotation, and version tags.
- [x] **Adding GitLab (PR #66) requires one new file and a registry entry — no
      changes to `event_router.py`.**
      `test_adding_a_provider_takes_a_registry_entry_and_nothing_else` registers
      a source the service has never heard of, with its own parser, a
      non-default verifier, its own signature header and a secret accessor, and
      routes a real HTTP delivery end to end. Nothing outside the descriptor was
      added to make it pass.
- [x] **`RESERVED_SOURCES` is derived, not hand-maintained.** `reserved_sources()`
      returns `frozenset(PROVIDERS)`. Validation calls `is_builtin_source()`
      rather than the snapshot, so a provider registered after import time is
      also protected — tested.

## One behaviour change to be aware of

A `signature_scheme` naming no registered verifier now returns **500**, not a
silent fallback to the default. Only reachable by editing the row directly,
since the API validates the scheme on write. Falling back would reject every
genuine delivery as a bad signature and present a misconfiguration as an
authentication failure.

## API surface

`signature_scheme` is added to `CustomWebhookCreate` (optional, defaults to
`hmac_sha256_hex`), `CustomWebhookUpdate` (optional), and
`CustomWebhookResponse` / `CustomWebhookCreateResponse` (required). Verified
against the generated OpenAPI spec: those four schemas and no others.
`EventResponse` is unchanged, so phase 1's wire guarantee still holds.

## Verification

- Full suite on top of `main`: **1439 passed** (1361 on `main` alone, +78 new,
  all in `tests/test_providers.py`). No existing test modified except the two
  order-dependent ones described above.
- `pre-commit` (ruff format, ruff lint, pycodestyle, pyright) clean on all
  eleven files.
- Migration `018` applied and rolled back on **SQLite** (`upgrade head` →
  `downgrade -1`, column added nullable with server default, dropped cleanly)
  and on **real Postgres 15**, with a pre-existing row present across both
  directions.
- Renumbered `017` → `018` during the rebase onto `main`: #334 landed revision
  `017` in the meantime, and two files declaring `017` off `016` left alembic
  with two heads, refusing to run at all. The DDL is unchanged.

## Out of scope

Stream transports (#360), any use of `subject` (#362), and GitLab itself (#66) —
this phase only widens the description of a provider.

### DIFF
--- a/openhands/automation/event_schemas/__init__.py
+++ b/openhands/automation/event_schemas/__init__.py
@@ -13,7 +13,6 @@ expressions against the raw payload. The event schemas are for validation
 and providing typed access to payload fields.
 """
 
-from collections.abc import Callable
 from typing import Any, ClassVar
 
 from pydantic import BaseModel, computed_field
@@ -54,19 +53,13 @@ class WebhookEvent(BaseModel):
 
 
 # =============================================================================
-# Parser Registry
+# Parsing
 # =============================================================================
-
-# Type for auto-detection parse functions: (payload) -> WebhookEvent
-AutoParseFunc = Callable[[dict[str, Any]], WebhookEvent]
-
-# Registry of parse functions for known sources (auto-detect event type from payload)
-_PARSERS: dict[str, AutoParseFunc] = {}
-
-
-def register_parser(source: str, parser: AutoParseFunc) -> None:
-    """Register a parse function for a source."""
-    _PARSERS[source] = parser
+#
+# The parser for a known source lives on its provider descriptor
+# (`openhands.automation.providers`), alongside its verifier, secret and
+# capabilities, so a provider is described in one place. `parse_event()` reads
+# that registry; there is no separate parser registry here.
 
 
 def parse_event(
@@ -94,10 +87,14 @@ def parse_event(
     Raises:
         ValueError: If event type cannot be determined from payload
     """
+    # Imported lazily: providers.py imports the parsers this module's
+    # submodules define, so a module-level import here would be circular.
+    from openhands.automation.providers import get_provider
+
     # Known source - auto-detect event type from payload
-    parser = _PARSERS.get(source)
-    if parser:
-        return parser(payload)
+    provider = get_provider(source)
+    if provider:
+        return provider.parse(payload)
 
     # Unknown source = custom webhook (no registration needed)
     from openhands.automation.event_schemas.custom import (
@@ -114,19 +111,3 @@ def parse_event(
         payload=payload,
         source_override=source,
     )
-
-
-def _register_builtin_parsers() -> None:
-    """Register parsers for built-in sources. Called at module load."""
-    from openhands.automation.event_schemas.bitbucket_data_center import (
-        parse_bitbucket_data_center_event,
-    )
-    from openhands.automation.event_schemas.github import parse_github_event_auto
-    from openhands.automation.event_schemas.jira_dc import parse_jira_dc_event
-
-    register_parser("bitbucket_data_center", parse_bitbucket_data_center_event)
-    register_parser("github", parse_github_event_auto)
-    register_parser("jira_dc", parse_jira_dc_event)
-
-
-_register_builtin_parsers()
--- a/openhands/automation/event_schemas/__init__.py
+++ b/openhands/automation/event_schemas/__init__.py
@@ -52,16 +52,6 @@ class WebhookEvent(BaseModel):
         raise NotImplementedError("Subclasses must implement event_key")
 
 
-# =============================================================================
-# Parsing
-# =============================================================================
-#
-# The parser for a known source lives on its provider descriptor
-# (`openhands.automation.providers`), alongside its verifier, secret and
-# capabilities, so a provider is described in one place. `parse_event()` reads
-# that registry; there is no separate parser registry here.
-
-
 def parse_event(
     source: str,
     payload: dict[str, Any],
@@ -87,8 +77,7 @@ def parse_event(
     Raises:
         ValueError: If event type cannot be determined from payload
     """
-    # Imported lazily: providers.py imports the parsers this module's
-    # submodules define, so a module-level import here would be circular.
+    # Lazy: providers imports WebhookEvent from this module.
     from openhands.automation.providers import get_provider
 
     # Known source - auto-detect event type from payload
--- a/openhands/automation/providers.py
+++ b/openhands/automation/providers.py
@@ -1,24 +1,10 @@
 """Provider descriptors and signature verifiers.
 
-One registry describes an event provider: how its payloads parse into a typed
-`WebhookEvent`, how a delivery's signature is verified, where its shared secret
-comes from, and what its transport tolerates. Before this module the same
-provider was described in two unrelated places -- ``BUILTIN_SOURCES`` in
-``utils/webhook.py`` (secret extraction) and the ``event_schemas`` parser
-registry -- with ``RESERVED_SOURCES`` hand-maintained alongside them as a third.
-
-Verification is a registry too. ``verify_signature()`` was the only scheme the
-service could speak: hex HMAC-SHA256 over the raw body. Providers that sign
-something else -- Standard Webhooks (GitLab 19.1+, Svix), Slack's
-``v0:{ts}:{body}`` -- could not be onboarded at all. ``VERIFIERS`` resolves a
-scheme by name so a provider, or a per-org custom webhook, can pick one.
-
-This module is deliberately free of transport imports so it can be read by a
-non-HTTP transport; the one HTTP-shaped field (``handshake``) is reserved and
-typed under ``TYPE_CHECKING`` only.
+`PROVIDERS` describes an event provider in one place: how its payloads parse,
+how a delivery is verified, where its secret comes from, what its transport
+tolerates. `VERIFIERS` resolves a signature scheme by name, so a provider or a
+per-org custom webhook can pick one. No transport imports.
 """
-
-from __future__ import annotations
 
 import base64
 import binascii
@@ -28,7 +14,7 @@ import logging
 import time
 from collections.abc import Callable, Mapping
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any, Protocol
+from typing import TYPE_CHECKING, Any, Final, Protocol
 
 from openhands.automation.config import Settings
 from openhands.automation.event_schemas import WebhookEvent
@@ -43,32 +29,25 @@ if TYPE_CHECKING:
 logger = logging.getLogger("automation.providers")
 
 
+DEFAULT_VERIFIER: Final[str] = "hmac_sha256_hex"
+DEFAULT_BUILTIN_SIGNATURE_HEADER: Final[str] = "X-Hub-Signature-256"
+
+STANDARD_WEBHOOKS_TOLERANCE_SECONDS: Final[int] = 300
+STANDARD_WEBHOOKS_ID_HEADER: Final[str] = "webhook-id"
+STANDARD_WEBHOOKS_TIMESTAMP_HEADER: Final[str] = "webhook-timestamp"
+
+SLACK_TOLERANCE_SECONDS: Final[int] = 300
+SLACK_TIMESTAMP_HEADER: Final[str] = "X-Slack-Request-Timestamp"
+
+
 ParseFunc = Callable[[dict[str, Any]], WebhookEvent]
 SecretFunc = Callable[[Settings], str | None]
 SubjectFunc = Callable[[dict[str, Any]], "EventSubject | None"]
 HandshakeFunc = Callable[["Request"], "Response | None"]
 
-# Scheme used when nothing says otherwise: the behaviour every existing source
-# and every existing custom_webhooks row has today.
-DEFAULT_VERIFIER = "hmac_sha256_hex"
-
-# The header built-in sources sign into. They are all forwarded by the OpenHands
-# server, which reuses GitHub's header name regardless of the upstream provider.
-DEFAULT_BUILTIN_SIGNATURE_HEADER = "X-Hub-Signature-256"
-
-
-# =============================================================================
-# Header access
-# =============================================================================
-
 
 def get_header(headers: Mapping[str, str], name: str) -> str | None:
-    """Look up a header case-insensitively.
-
-    Starlette's ``Headers`` is already case-insensitive, but a non-HTTP caller
-    or a test may pass a plain dict, and header casing is not something a
-    verifier should have to guess at.
-    """
+    """Look up a header case-insensitively, for callers passing a plain dict."""
     value = headers.get(name)
     if value is not None:
         return value
@@ -86,20 +65,12 @@ def _now_seconds() -> int:
     return int(time.time())
 
 
-# =============================================================================
-# Verifiers
-# =============================================================================
-
-
 class WebhookVerifier(Protocol):
     """Proves a raw delivery was signed by the holder of the shared secret.
 
-    ``signature_header`` is the header the signature itself is carried in. It
-    is a parameter rather than a property of the verifier because custom
-    webhooks configure it per row, and the same scheme is used behind different
-    header names by different vendors. Everything *else* a scheme needs -- a
-    message id, a timestamp -- the verifier reads from ``headers`` itself,
-    since those header names are fixed by the scheme, not by the operator.
+    `signature_header` is a parameter because custom webhooks configure it per
+    row; anything else a scheme needs is read from `headers`, since those names
+    are fixed by the scheme.
     """
 
     def verify(
@@ -113,22 +84,7 @@ class WebhookVerifier(Protocol):
 
 
 def verify_signature(payload: bytes, signature: str, secret: str) -> bool:
-    """
-    Verify HMAC-SHA256 signature.
-
-    Accepts both formats:
-    - GitHub/normalized: 'sha256=<hex>'
-    - Raw hex digest: '<hex>' (e.g., Linear)
-
-    Args:
-        payload: Raw request body bytes
-        signature: Signature from header
-        secret: The shared secret
-
-    Returns:
-        True if signature is valid
-    """
-    # Normalize: strip 'sha256=' prefix if present
+    """Verify a hex HMAC-SHA256 digest, with or without a `sha256=` prefix."""
     if signature.startswith("sha256="):
         signature = signature[7:]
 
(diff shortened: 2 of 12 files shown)


openhands/automation/event_schemas/__init__.py
```diff
@@ -54,19 +53,13 @@ def event_key(self) -> str:
...
-AutoParseFunc = Callable[[dict[str, Any]], WebhookEvent]
-
-# Registry of parse functions for known sources (auto-detect event type from payload)
-_PARSERS: dict[str, AutoParseFunc] = {}
-
-
-def register_parser(source: str, parser: AutoParseFunc) -> None:
-    """Register a parse function for a source."""
-    _PARSERS[source] = parser
+#
```
VascoSch92: delete this comment

openhands/automation/event_schemas/__init__.py
```diff
@@ -94,10 +87,14 @@ def parse_event(
     Raises:
         ValueError: If event type cannot be determined from payload
     """
+    # Imported lazily: providers.py imports the parsers this module's
+    # submodules define, so a module-level import here would be circular.
+    from openhands.automation.providers import get_provider
```
VascoSch92, line 81: are you sure about that? Can you double check?

openhands/automation/utils/webhook.py
```diff
@@ -1,98 +1,55 @@
...
+    "create_automation_run",
+    "get_event_automations",
+    "get_requested_event_types",
+    "get_webhook_config",
+    "is_builtin_source",
+    "verify_signature",
+]
 
 
 # =============================================================================
```
VascoSch92: delete this comment. And in general I don't want this kind of comments

openhands/automation/providers.py
```diff
@@ -0,0 +1,391 @@
...
+something else -- Standard Webhooks (GitLab 19.1+, Svix), Slack's
+``v0:{ts}:{body}`` -- could not be onboarded at all. ``VERIFIERS`` resolves a
+scheme by name so a provider, or a per-org custom webhook, can pick one.
+
+This module is deliberately free of transport imports so it can be read by a
+non-HTTP transport; the one HTTP-shaped field (``handshake``) is reserved and
+typed under ``TYPE_CHECKING`` only.
+"""
+
+from __future__ import annotations
```
VascoSch92: do we need that?

openhands/automation/providers.py
```diff
@@ -0,0 +1,391 @@
...
+HandshakeFunc = Callable[["Request"], "Response | None"]
+
+# Scheme used when nothing says otherwise: the behaviour every existing source
+# and every existing custom_webhooks row has today.
+DEFAULT_VERIFIER = "hmac_sha256_hex"
+
+# The header built-in sources sign into. They are all forwarded by the OpenHands
+# server, which reuses GitHub's header name regardless of the upstream provider.
+DEFAULT_BUILTIN_SIGNATURE_HEADER = "X-Hub-Signature-256"
+
```
VascoSch92: Use FInal[str] here

openhands/automation/providers.py
```diff
@@ -0,0 +1,391 @@
...
+
+# =============================================================================
```
VascoSch92: delete this comment

openhands/automation/providers.py
```diff
@@ -0,0 +1,391 @@
...
+        if key.lower() == lowered:
+            return candidate
+    return None
+
+
+def _now_seconds() -> int:
+    return int(time.time())
+
+
+# =============================================================================
```
VascoSch92: delete

openhands/automation/providers.py
```diff
@@ -0,0 +1,391 @@
...
+        return verify_signature(body, signature, secret)
+
+
+# Standard Webhooks (https://www.standardwebhooks.com) -- used by GitLab 19.1+
+# signing tokens, Svix, and others. Signature = base64(HMAC-SHA256(key, msg))
+# where msg = "{id}.{timestamp}.{body}" and key = base64decode(secret w/o whsec_).
+STANDARD_WEBHOOKS_TOLERANCE_SECONDS = 300
+STANDARD_WEBHOOKS_ID_HEADER = "webhook-id"
+STANDARD_WEBHOOKS_TIMESTAMP_HEADER = "webhook-timestamp"
+
```
VascoSch92: usa Final with the correct typing here

openhands/automation/providers.py
```diff
@@ -0,0 +1,391 @@
...
+                continue
+            if hmac.compare_digest(candidate, expected):
+                matched = True
+        return matched
+
+
+# Slack's Events API signs "v0:{timestamp}:{body}" and sends the hex digest as
+# "v0=<hex>". Slack's own guidance is to reject deliveries older than 5 minutes.
+SLACK_TOLERANCE_SECONDS = 300
+SLACK_TIMESTAMP_HEADER = "X-Slack-Request-Timestamp"
```
VascoSch92: as above. Moreover, port all these contast at the beginning of the file

openhands/automation/providers.py
```diff
@@ -0,0 +1,391 @@
...
+    """Resolve a signature scheme name, treating None as the default."""
+    return VERIFIERS.get(scheme or DEFAULT_VERIFIER)
+
+
+def verifier_schemes() -> frozenset[str]:
+    """Every signature scheme a webhook may be configured with."""
+    return frozenset(VERIFIERS)
+
+
+# =============================================================================
```
VascoSch92: delete

openhands/automation/providers.py
```diff
@@ -0,0 +1,391 @@
...
+    source: str
+    parse: ParseFunc
+    verifier: str = DEFAULT_VERIFIER
+    signature_header: str = DEFAULT_BUILTIN_SIGNATURE_HEADER
+    secret_from_settings: SecretFunc | None = None
+    # Reserved for #362 (subject -> conversation routing). Nothing reads it.
+    subject: SubjectFunc | None = None
+    # Reserved. Deliberately not wired: no provider on the roadmap performs an
+    # HTTP handshake. Slack's Events API challenge was the motivating example,
+    # and #360 makes Socket Mode our Slack path, where it does not apply. See
```
VascoSch92: we don't want issue number here

openhands/automation/providers.py
```diff
@@ -0,0 +1,391 @@
...
+    # the "On the handshake hook" note in #359 -- this is the explicit decision
```
VascoSch92: als here no issue number

openhands/automation/providers.py
```diff
@@ -0,0 +1,312 @@
...
+        expected = (
+            "v0="
+            + hmac.new(
+                secret.encode("utf-8"), signed_content, hashlib.sha256
+            ).hexdigest()
+        )
+        return hmac.compare_digest(signature, expected)
+
+
+VERIFIERS: dict[str, WebhookVerifier] = {
```
VascoSch92: this is also on top

malhotra5 (approved):

PR #381, opened 2026-08-24
feat: persist accepted events to deduplicate redeliveries and expose events that matched nothing

### LINKED ISSUE #361: Persist accepted events to deduplicate redeliveries and expose events that matched nothing
> **Phase 4 of 5** · part of #363
>
> **Blocked by:** #358 — hard. The dedupe check lives inside `accept_event()`.
> **Blocks:** nothing directly, but #360 is only safe at a single replica until this lands.
>
> Schema change — the first irreversible step in the sequence.

## Background

`receive_event()` today verifies, parses, queries automations, matches triggers,
inserts runs, commits, and responds — all synchronously, and **without ever
persisting the event itself.** The only durable trace of an incoming event is the
`AutomationRun` rows it happened to create.

If it created none, there is no trace at all.

## Problem

Six consequences, in rough order of how often they bite:

1. **No provider-event deduplication.** GitHub sends delivery IDs, Slack sends
   `event_id`, most providers send something. We use none of them. A redelivery
   starts a duplicate run.
2. **Replay is unsafe.** Signatures do not expire (noted in the module docstring
   at `event_router.py:25–29`), so a captured payload can be replayed to start the
   same run again. The current comment argues this is acceptable because triggers
   are idempotent, which is true of some automations and not others.
3. **Unmatched events are invisible.** A user whose JMESPath filter is subtly
   wrong sees silence, and has no way to distinguish "the event never arrived"
   from "it arrived and matched nothing". This is the single most common support
   question shape for event triggers.
4. **No retry if routing fails** after the provider already considers delivery
   successful.
5. **No shared delivery record** across transports — each new transport would
   otherwise invent its own reliability story.
6. **Weak debugging.** Runs are visible; the events that did or did not produce
   them are not.

## Proposed change

An `IntegrationEvent` row written **in the same transaction as the runs**, inside
`accept_event()`:

```
IntegrationEvent
  id, org_id, source, provider_event_id, event_key,
  payload, received_at, matched_count, processing_status, error

  PARTIAL UNIQUE (source, provider_event_id) WHERE provider_event_id IS NOT NULL
```

The partial index matters: not every provider supplies a delivery ID, and custom
webhooks may supply none. Those events are still recorded, just not deduplicated.

On unique-violation, `accept_event()` returns `AcceptResult(duplicate=True)` and
creates nothing. The HTTP handler still answers 2XX — a duplicate is a successful
delivery from the provider's point of view, and answering anything else invites
further retries.

Providers populate `AcceptedEvent.provider_event_id` (phase 1 already reserved the
field): GitHub's `X-GitHub-Delivery`, Slack's `event_id`, and so on.

Add pruning to the existing watchdog loop — nothing currently prunes, and this
table grows with every delivery.

## Explicitly out of scope: the async routing worker

Earlier discussion paired this table with a background worker that would do the
matching and run creation, leaving the receiver to persist-and-return.

That is a separate change and should not ride along with this one. It solves
"slow routing blocks the HTTP response", which we do not have: routing is one
indexed query plus a JMESPath evaluation plus N inserts, against GitHub's
documented budget of *"a 2XX response within 10 seconds… if your server takes
longer, GitHub terminates the connection and considers the delivery a failure"*.
We are orders of magnitude inside that.

It also costs a queue, a worker loop, added latency, and a new class of failure —
and the reversibility is asymmetric: **deferring the worker is cheap to reverse,
building it is not.** Async routing can be added later behind an unchanged
`accept_event()` signature; removing a worker that automations have come to
depend on for timing is much harder.

Concrete revisit condition, so this is settled by measurement rather than opinion:
instrument routing latency in the handler and revisit if p99 approaches ~1s.

## Why this is required, not cosmetic

The VM bridge deduplicates in a process-local dict, which is correct for a single
process. Once stream sources run in-service (phase 3) across more than one
replica, Slack's redelivery of an un-acked envelope can land on a different
connection in a different process, where that in-memory set is invisible.

**Persistent deduplication is what makes multi-replica stream sources correct**,
and multi-replica is something Slack explicitly supports (up to 10 connections,
payloads distributed across them). Without this phase, phase 3 is safe only at
one replica.

## Acceptance criteria

- [ ] The same GitHub delivery ID posted twice creates runs once and returns 2XX
      both times.
- [ ] The same Slack `event_id` delivered twice — including via two different
      connections in two processes against a shared database — creates runs once.
- [ ] An event matching no automation is recorded with `matched_count = 0` and is
      queryable.
- [ ] Events without a `provider_event_id` are still recorded, and are not
      deduplicated against each other.
- [ ] The event row and its runs commit atomically: a failure creating runs
      leaves no event row claiming success.
- [ ] Pruning is in place and bounded.

## Out of scope

Async routing (see above). Any UI for browsing events — this phase makes the data
exist; surfacing it is a follow-up.

VascoSch92:
Closes #361 — phase 4 of 5, part of #363.

**Stacked on #378** (phase 2). Base branch is `vasco/provider-descriptor-registry`;
review only the commit on top of it.

One new table, one new nullable-free column set, no behaviour change for any
event that is not a redelivery.

## What changed

`accept_event()` verified nothing about *the delivery itself*. It matched
triggers, created runs, committed, and forgot. If an event matched no
automation it created no runs, and so left no trace at all — "did my webhook
arrive, or is my filter wrong?" had no answer anywhere in the service.

`IntegrationEvent` is written inside `accept_event()`, in the same transaction
as the runs:

```
integration_events
  id, org_id, source, provider_event_id, event_key,
  payload, matched_count, received_at

  UNIQUE (org_id, source, provider_event_id) WHERE provider_event_id IS NOT NULL
```

The insert happens **first**, before the match pass, so it doubles as the
dedupe gate: a redelivery costs one failed INSERT instead of a full routing
pass. A unique violation rolls back to a savepoint — without which the failure
would poison the transaction the caller still has to commit — and returns
`AcceptResult(duplicate=True)`. The HTTP handler still answers 200: from the
provider's side the delivery succeeded, and anything else only invites more
retries.

Transports supply the id, the same way they supply authentication.
`Provider` gains `event_id_header`, and the HTTP receiver reads it
(`X-GitHub-Delivery` for GitHub). A stream transport is handed the envelope and
sets `AcceptedEvent.provider_event_id` from it directly — which is how #360 gets
multi-replica dedupe for free.

Pruning goes in the watchdog loop, bounded by `integration_event_retention_days`
(default 14) and by `PRUNE_BATCH_SIZE` per scan.

## Five decisions worth reviewing

**1. The dedupe key is scoped by org.**

The issue sketches `PARTIAL UNIQUE (source, provider_event_id)`. That is unsafe
for custom webhooks: `source` is only unique *per org*
(`ix_custom_webhooks_org_source`), so two orgs each running a webhook they both
call `ci`, each numbering deliveries from 1, would deduplicate against each
other — the second org's first event would vanish silently.

Adding `org_id` weakens nothing. A delivery belongs to exactly one org either
way, so GitHub and Slack dedupe identically. Test:
`test_accept_event_deduplicates_within_one_org_only`.

**2. `processing_status` and `error` are not in the table.**

The issue lists them. With the async routing worker explicitly out of scope,
neither column has a writer: routing is synchronous, so a failure to create runs
rolls the whole transaction back and there is no row to put an error on, and
`processing_status` would be one constant value on every row ever written.

This is the issue's own reversibility argument applied one level down — the
worker that gives those columns meaning will need its own migration anyway (a
claim/lease column at minimum), and adding two nullable columns then is cheaper
than carrying two dead ones now. Happy to add them if you'd rather have the
shape settled up front.

**3. The delivery id is a dedupe key, not replay protection.**

`X-GitHub-Delivery` is not covered by the signature, so a replayer who alters
it gets a fresh id. That is worth being explicit about, because problem #2 in
the issue is "replay is unsafe" and this only partly addresses it: verbatim
replays now no-op, crafted ones do not. Real replay protection is the
timestamp-signing schemes phase 2 added (`standard_webhooks`, `slack_v0`). The
module docstring says so rather than implying the hole is closed.

**Confirmed: built-in deliveries do not currently carry this header, so it is
inert for them today.** Built-in sources do not reach us straight from the
provider — they are re-enveloped and re-signed upstream, and that hop forwards
only the content type and the signature. I checked; no delivery id survives it.

So for now **every forwarded built-in event gets `provider_event_id = NULL`**:
recorded, routed, not deduplicated — today's behaviour plus the row.
`test_github_event_without_a_delivery_header_is_still_recorded` pins exactly
that. **Acceptance criterion 1 is therefore not met end-to-end by this PR
alone**, and a reviewer should read it that way: the mechanism here is complete
and tested, the producer upstream is missing. Tracked separately.

When that producer is added, a header is the wrong place to put it. The
signature covers only the body, so a header is tamperable, whereas the envelope
that hop already builds is signed. Reading the id out of the body is also what
a Slack envelope needs. Expect `event_id_header` to gain a body-expression
sibling rather than be replaced.

**4. A violation with no id is re-raised, not reported as a duplicate.**

The index is partial, so a row with `provider_event_id IS NULL` cannot conflict
on it. If an `IntegrityError` still arrives, something else failed, and
swallowing it as a duplicate would drop a genuine event without a trace.

**5. Pruning rides the watchdog rather than getting its own loop.**

It is the service's only periodic janitor. A second loop with its own interval,
task and shutdown handling buys nothing for one bounded DELETE.

## Two indexes, not three

The partial unique index and `received_at` (which pruning uses). An
`(org_id, received_at)` index is what a browse-events UI will want, and it can
land with the query surface that needs it — this phase makes the data exist.

## Acceptance criteria

| Criterion | Test |
|---|---|
| Same GitHub delivery ID twice → runs once, 2XX both times | `test_redelivered_github_event_creates_runs_once` |
| Same id via two processes against a shared DB → runs once | `test_accept_event_deduplicates_through_the_database` |
| Event matching nothing recorded with `matched_count = 0` | `test_accept_event_records_events_that_match_nothing` |
| Events without an id recorded, not deduplicated | `test_accept_event_does_not_deduplicate_without_an_id` |
| Event row and runs commit atomically | `test_accept_event_records_and_creates_runs_atomically` |
| Pruning in place and bounded | `TestPruneIntegrationEvents` |

Plus `test_accept_event_dedupe_survives_a_rolled_back_delivery`: a delivery that
failed to route can be retried under the same id, because the dedupe row rolls
back with everything else. Without that, one transient failure would swallow
the event permanently.

## Out of scope

Async routing worker (per the issue). Any UI for browsing events. Dedupe for
custom webhooks — the `standard_webhooks` scheme signs its `webhook-id`, which
would make it a *replay-proof* key rather than merely a dedupe one, and that is
a natural follow-up.

## Verification

- Full suite: 1454 passed (re-run after merging `main`).
- `pre-commit run --all-files`: clean (ruff format, ruff lint, pycodestyle, pyright).
- Migration `019` checked offline against PostgreSQL; the partial unique index
  emits as expected. SQLite covered by the existing `alembic upgrade head` test.
- **Deployed to the shared OSS automation VM** (`OH_AUTOMATION_GIT_REF`), which
  runs SQLite with 30 automations and ~109k `automation_runs`. The event-log
  migration applied on startup with no downtime beyond the normal restart, the
  table and partial index landed, and the stack came back healthy. A
  webhook-driven test automation there confirmed end to end that a matching
  event creates a run and is recorded, and that an event filtered out by
  JMESPath is recorded with `matched_count = 0`.
- That same VM test also demonstrated the gap above from the other direction: a
  custom-source webhook fed a real `X-GitHub-Delivery` header still recorded
  `provider_event_id = NULL` and created a second run on redelivery, because
  only a registered `Provider` supplies a delivery id. Custom webhooks having
  no dedupe is per #361, but it is worth seeing rather than assuming.

- Merged `main` in. Phases 1 and 2 are now squash-merged there (#367, #378), so
  the branch's own copies of those commits conflicted with their squashed
  equivalents; every conflicted file was byte-identical between the two sides,
  so the branch version stands. `main` also landed `017_add_run_metadata`, which
  pushed the webhook-signature migration to `018` — so this PR's migration is
  renumbered `019_add_integration_events` (`down_revision = "018"`). The chain
  is linear 016→019 with a single head, which `tests/test_db.py` pins by running
  a real `alembic upgrade head`.

HUMAN: Deployed this branch to the shared OSS automation VM and drove real webhooks
through it end to end, rather than trusting the test suite alone.

- A matching webhook creates a run and is recorded; an event filtered out by
  JMESPath is recorded with `matched_count = 0`.
- A custom-source webhook carrying an `X-GitHub-Delivery` header still recorded
  `provider_event_id = NULL` and created a second run on redelivery — the gap
  described above, observed rather than assumed.
- Migrations applied on startup against the VM's live SQLite database (30
  automations, ~109k `automation_runs`) with no downtime beyond the restart.

### DIFF
--- a/openhands/automation/ingest.py
+++ b/openhands/automation/ingest.py
@@ -1,22 +1,8 @@
-"""
-Transport-neutral event ingestion.
+"""Transport-neutral event ingestion.
 
-`accept_event()` is the boundary between *how an event arrived* and *what the
-service does with it*. Everything above the boundary — reading bytes, proving
-the event is genuine, turning a provider payload into an ``event_key`` — belongs
-to the transport. Everything below it — finding the org's event automations,
-matching triggers, creating PENDING runs — is identical for every transport and
-lives here.
-
-Authentication is deliberately **not** part of this module. HTTP proves
-authenticity with an HMAC over the raw body; a Socket Mode connection proves it
-by possession of an app-level token over TLS; a poller proves it by having made
-the outbound call itself. `accept_event()` receives an already-authenticated
-event and never asks how it was authenticated. Without that rule, every
-non-HTTP transport is tempted to manufacture a signature to satisfy a check that
-does not apply to it.
-
-The only caller today is the webhook handler in `event_router.py`.
+`accept_event()` separates how an event arrived from what the service does with
+it. Transports own acquisition, authentication and interpretation; trigger
+matching and run creation live here and are shared by every transport.
 """
 
 import logging
@@ -40,40 +26,29 @@ from openhands.automation.utils.webhook import (
 logger = logging.getLogger("automation.ingest")
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
 class EventSubject:
-    """The external thing an event is about (a Slack thread, a pull request).
-
-    Reserved for subject -> conversation routing; nothing populates or reads it
-    yet. `key` is the value that will be stored as `subject_key`.
-    """
+    """The external thing an event is about. Reserved; nothing reads it yet."""
 
     key: str
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
 class AcceptedEvent:
-    """An authenticated, interpreted event, ready to be routed.
-
-    The transport owns everything that produces one of these: acquiring the
-    bytes, proving they are genuine, and deriving `event_key`.
-    """
+    """An authenticated, interpreted event, ready to be routed."""
 
     source: str
     event_key: str
-    # Raw provider payload. JMESPath trigger filters run against this.
+    # Raw provider payload; JMESPath trigger filters run on this.
     payload: dict[str, Any] = field(default_factory=dict)
     provider_event_id: str | None = None
     subject: EventSubject | None = None
     occurred_at: datetime | None = None
-    # Optional typed event, when the transport parsed the payload into a
-    # Pydantic model (the webhook handler does this via `parse_event()`). When
-    # present it — not `payload` — is what gets persisted as the run's
-    # `event_payload`, preserving the shape existing automations already see.
+    # When set, persisted as the run's event_payload in place of `payload`.
     parsed_event: BaseModel | None = None
 
 
-@dataclass
+@dataclass(frozen=True, slots=True)
 class AcceptResult:
     """The outcome of routing an accepted event."""
 
@@ -89,29 +64,18 @@ async def accept_event(
     *,
     request: Request | None = None,
 ) -> AcceptResult:
-    """
-    Route an already-authenticated event to the org's matching automations.
+    """Route an already-authenticated event to the org's matching automations.
 
-    Args:
-        org_id: The organization the event belongs to
-        event: The authenticated, interpreted event
-        session: Database session (committed before returning)
-        request: Originating HTTP request, for telemetry only. Non-HTTP
-                 transports omit it; every telemetry event still fires.
-
-    Returns:
-        AcceptResult with the number of matched automations and the IDs of the
-        runs created for them.
+    Commits before returning. `request` is telemetry only; non-HTTP transports
+    omit it and every telemetry event still fires.
     """
     source = event.source
     webhook_payload = event.payload
 
-    # 1. Find matching automations
     automations = await get_event_automations(org_id, source, session)
     matched_automations = []
 
     for automation, trigger in automations:
-        # Match trigger against webhook payload using JMESPath filter
         if matches_trigger(trigger, source, event.event_key, webhook_payload):
             matched_automations.append(automation)
 
@@ -133,9 +97,7 @@ async def accept_event(
         },
     )
 
-    # 2. Create PENDING runs for matched automations
-    # For Pydantic-parsed events (GitHub), use model_dump() for typed fields
-    # For custom webhooks, use the webhook payload directly
+    # Typed events (GitHub) keep their model shape; others store the payload.
     event_payload = (
         event.parsed_event.model_dump(mode="json")
         if isinstance(event.parsed_event, BaseModel)
--- a/openhands/automation/ingest.py
+++ b/openhands/automation/ingest.py
@@ -13,7 +13,7 @@ from typing import Any
 
 from fastapi import Request
 from pydantic import BaseModel
-from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.ext.asyncio import AsyncSession, async_sessionmaker
 
 from openhands.automation.telemetry import capture_automation_event
 from openhands.automation.trigger_matcher import matches_trigger
@@ -63,11 +63,19 @@ async def accept_event(
     session: AsyncSession,
     *,
     request: Request | None = None,
+    session_factory: async_sessionmaker[AsyncSession] | None = None,
 ) -> AcceptResult:
     """Route an already-authenticated event to the org's matching automations.
 
-    Commits before returning. `request` is telemetry only; non-HTTP transports
-    omit it and every telemetry event still fires.
+    Commits before returning.
+
+    `request` and `session_factory` are both telemetry plumbing. Telemetry
+    resolves its distinct id from the database, and HTTP callers supply that
+    reader indirectly as `request.app.state.session_factory`; a caller that
+    passes neither drops every event silently. So non-HTTP transports must pass
+    `session_factory`. It is deliberately not `session`: telemetry writes its id
+    row, and sharing this session would move that write into the caller's
+    transaction.
     """
     source = event.source
     webhook_payload = event.payload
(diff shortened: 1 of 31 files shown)


openhands/automation/ingest.py
```diff
@@ -80,13 +84,44 @@ async def accept_event(
...
+    # Insert before doing any routing work, so a redelivery costs one failed
+    # INSERT rather than a match pass. The savepoint is what keeps the
+    # violation recoverable: without it the failure would poison the
+    # transaction the caller still has to commit.
+    try:
+        async with session.begin_nested():
+            session.add(record)
+            await session.flush()
+    except IntegrityError:
+        if event.provider_event_id is None:
```
malhotra5, line 103: NIT: we can avoid having none type for the provider event id by generating an id when unspecified. a quick solution would be to hash the json payload to generate an id

VascoSch92: I think it is already the whole traffic today that... So should we change that? 

Not a problem just to know :-)

malhotra5: Hmm fair

We can't rely on hashing just a fixed length substring as that's not reliable, so yep okay with not doing this

malhotra5 (approved):

PR #384, opened 2026-08-25
feat: add in-service Slack Socket Mode via a supervised stream transport

### LINKED ISSUE #360: Add in-service Slack Socket Mode via a supervised stream transport

VascoSch92:
Closes #360. Phase 3 of #363.

Slack Socket Mode already works in production — as `slack-socket-bot.service`, a separate systemd unit on the OSS VM that acks Slack and then POSTs the envelope back to this service's own `/v1/events/{org}/slack` endpoint on a daemon thread with no retry. Two forwards have timed out in eight days, and because the envelope was already acked, Slack never redelivered them. In-process there is no POST, no timeout and no thread, so that failure mode ceases to exist.

## What this adds

**`openhands/automation/streams/`**

- `base.py` — the `StreamProvider` protocol. `run(emit, shutdown)` rather than `connect`/`receive`/`close`: `slack_sdk` already owns a correct reconnect loop, and forcing it into a poll shape means reimplementing that badly. Also the per-source health record (`last_connected_at`, `last_event_at`, `consecutive_failures`), kept process-local.
- `supervisor.py` — the source registry (fed from settings, no table) and one supervised child task per source. Exceptions are caught **per source**, so one bad provider cannot take down the others, the scheduler, the dispatcher, or HTTP webhook handling — this is what buys back the isolation otherwise lost by leaving systemd. Restarts use exponential backoff; a `StreamConfigError` is terminal, since nothing a restart can fix.
- `slack.py` — one Socket Mode connection per Slack app, porting the four semantics from the VM script:
  1. **ack first**, before any other work
  2. **drop bot messages** (`bot_id`, or `subtype == "bot_message"`)
  3. **assert `team_id`** on every envelope
  4. **assert identity at startup** — `auth.test` must return both the expected `team_id` and `bot_user_id`, or the source refuses to start

  Uses the async `socket_mode.aiohttp` client, not the VM's threaded `builtin` one — the thread only existed to marshal a blocking POST off the socket, and there is no POST any more.

Dedupe is the one behaviour that did **not** need porting: #381 landed durable dedupe on `provider_event_id`, so the provider passes Slack's `event_id` to `accept_event()` and a redelivered envelope creates no second run, across replicas and restarts. No bounded LRU.

**Wiring** — `app.py` starts the supervisor in the lifespan alongside the existing loops and joins it on shutdown.

**Configuration** — `AUTOMATION_SLACK_APPS`, a JSON list of `{org_id, app_token, bot_token, team_id, bot_user_id}`. Configuring one is the whole switch: `enabled` requires a source, so a deployment that configures none starts nothing. `AUTOMATION_STREAMS_ENABLED` defaults **true** and exists only as a kill switch, for holding the sockets down without unsetting credentials. Deliberately no `StreamSource` table and no CRUD/status API: per-org socket configuration is a multi-tenant requirement, and taking it now costs a migration and a credential-encryption surface before anyone needs it.

**Dependency** — adds `slack-sdk` and `aiohttp`; neither was a dependency before.

## Acceptance criteria

- [x] With no `AUTOMATION_SLACK_APPS` configured (the default), behaviour is byte-identical to today — the supervisor never starts.
- [x] An `app_mention` creates the same runs from the **same unmodified automation definitions**: the payload handed to `accept_event()` is the whole Slack envelope, which is what the bridge POSTs today and what existing triggers filter on (`team_id == '...'` reads a top-level field).
- [x] Killing one source's connection does not affect the other source or any other loop.
- [x] A redelivered envelope creates no second run.
- [x] A wrong `team_id`/`bot_user_id` refuses to start that source and logs why.
- [x] Shutdown closes connections cleanly inside the existing 5s join.

## Testing

18 new tests in `tests/test_streams.py` covering the configuration gate and its kill switch, env-driven source construction, ack ordering, every drop rule, both identity mismatches, source isolation, non-retry of a misconfigured source, clean shutdown, and an end-to-end emit → run → redelivery-drops case against the database.

`1472 passed` on the full unit suite; `pre-commit` (ruff format, ruff lint, pycodestyle, pyright) clean.

## Cutover

Deploy with the flag on and the Slack app config, verify runs are created from in-service events, then `systemctl disable --now slack-socket-bot` on the OSS VM. No automation definitions change. Roll back by re-enabling the unit; both can run briefly in parallel, since Slack load-balances across connections and #381's persistent dedupe covers the overlap.

## Out of scope

The generic WebSocket source from #134, DB-backed stream sources with CRUD, any Canvas UI, threading/conversation reuse (phase 5), and event types beyond `app_mention`.


openhands/automation/streams/base.py
```diff
@@ -0,0 +1,77 @@
...
+    def name(self) -> str:
+        """Identifies this connection in logs and health; a source may have many."""
+        ...
+
+    async def run(self, emit: Emit, shutdown: asyncio.Event) -> None:
+        """Hold the connection until `shutdown` is set, emitting as events land."""
+        ...
+
+
+@dataclass
```
VascoSch92: slots=True and frozen

openhands/automation/streams/__init__.py
```diff
@@ -0,0 +1,34 @@
+"""Stream sources: inbound events over a held-open connection rather than HTTP.
+
+Submodules:
+    base.py        StreamProvider protocol, per-source health
+    slack.py       Slack Socket Mode provider
+    supervisor.py  Source registry and the supervised task per source
+
+Off unless `AUTOMATION_STREAMS_ENABLED` is set. Everything past `emit()` is the
```
VascoSch92: Why it is off unless? What is the idea behind?

openhands/automation/streams/slack.py
```diff
@@ -0,0 +1,167 @@
+"""Slack Socket Mode as a stream source.
+
+Ports the semantics of the OSS VM's `slack-socket-bot` bridge, which has run
+this way in production since 2026-08-13. What changes is only where the code
```
VascoSch92: Don't speak about the OSS VM please

openhands/automation/streams/slack.py
```diff
@@ -0,0 +1,167 @@
...
+    Emit,
+    StreamConfigError,
+    health_for,
+)
+from openhands.automation.utils.time import utcnow
+
+
+logger = logging.getLogger("automation.streams.slack")
+
+SOURCE = "slack"
```
VascoSch92: Where do we use this ?

openhands/automation/streams/supervisor.py
```diff
@@ -0,0 +1,159 @@
...
+        delay = min(
+            settings.stream_backoff_seconds
+            * 2 ** min(max(health.consecutive_failures - 1, 0), _MAX_BACKOFF_DOUBLINGS),
+            settings.stream_max_backoff_seconds,
+        )
+        logger.info("Restarting stream source %s in %.0fs", provider.name, delay)
+        try:
+            await asyncio.wait_for(shutdown_event.wait(), timeout=delay)
+        except TimeoutError:
+            continue
```
VascoSch92: from contextlib use Suppress

openhands/automation/config.py
```diff
@@ -402,6 +403,60 @@ def enabled(self) -> bool:
         return bool(self.git_sync_repo_url and self.git_sync_enabled)
 
 
+# ---------------------------------------------------------------------------
+# StreamSettings - Stream source configuration
+# ---------------------------------------------------------------------------
+
```
VascoSch92: Delete this comment

pyproject.toml
```diff
@@ -30,6 +32,7 @@ dependencies = [
   "pyjwt>=2.8",
   "python-json-logger>=3",
   "pyyaml>=6",
+  "slack-sdk>=3.33",
```
VascoSch92, line 39: do we need to add slack-sdk? Why?

pyproject.toml
```diff
@@ -10,6 +10,8 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
+  # Socket Mode transport for slack-sdk's async client.
+  "aiohttp>=3.10",
```
VascoSch92, line 16: do we have xhttp here? or we need this onw?

AGENTS.md
```diff
@@ -377,6 +381,54 @@ agent server, which doesn't make sense for the multi-tenant SaaS deployment.
...
+  triggers filter on (`team_id == '...'`).
+- **Health** is process-local (`stream_health()`): `last_connected_at`,
+  `last_event_at`, `consecutive_failures` per connection. Nothing renders it
+  yet.
+- Slack permits up to 10 concurrent Socket Mode connections per app and
+  distributes payloads across them, so this is safe active-active and a rolling
+  deploy can drop nothing. That is a Slack guarantee, not a general one — see
+  `Capabilities.tolerates_multiple_connections` in `providers.py`.
+
+Cutover on the OSS VM: deploy with the flag on, verify runs are created from
```
VascoSch92: don't speak about the OSS VM

AGENTS.md
```diff
@@ -377,6 +381,54 @@ agent server, which doesn't make sense for the multi-tenant SaaS deployment.
   when one is already running — that cycle covers everything the new one
   would, and the lock would only queue it behind anyway.
 
+## Stream Sources
```
VascoSch92, line 384: Ok make it more coincise.

openhands/automation/streams/base.py
VascoSch92: Done — `SourceHealth` is now `@dataclass(frozen=True, slots=True)`, matching the rest of the codebase. Since the supervisor has to update it, writes go through a new `record_health(name, **changes)` that stores a `replace()`d copy. Added tests for the frozen/slots contract and for a partial update leaving the other fields alone.

openhands/automation/streams/__init__.py
VascoSch92: Fair — it stated the rule without the reason. The docstring now says it: this is a self-hosted capability for two independent reasons. Slack does not allow Socket Mode apps in its public Marketplace, and a connection pinned to one replica does not fit a stateless autoscaled tier the way a request does. Webhooks stay the cloud path, and a deployment that sets nothing keeps exactly today's behaviour.

openhands/automation/streams/slack.py
VascoSch92: Removed. The module docstring now explains the design on its own terms — Slack holds the connection open and pushes envelopes down it, so the app token is the credential and the four rules are what keep the stream honest. I also scrubbed the other deployment-specific references in `supervisor.py` and in `accepted_event()`.

openhands/automation/streams/slack.py
VascoSch92: Only once, as the default for the `source` field right below it. Not worth the indirection — deleted, and the field is now `source: str = "slack"`.

openhands/automation/streams/supervisor.py
VascoSch92: Done. The `continue`/`return` pair was doing work the loop condition already does, so it collapses cleanly:

```python
# The loop condition re-reads the event, so a shutdown landing during
# the wait ends the source instead of restarting it.
with suppress(TimeoutError):
    await asyncio.wait_for(shutdown_event.wait(), timeout=delay)
```

openhands/automation/config.py
VascoSch92: Deleted. I had copied the banner from the neighbouring sections, but the class docstring already says what it is.

pyproject.toml
VascoSch92: Yes — it is the Socket Mode client itself. Socket Mode is not "a WebSocket you connect to": you call `apps.connections.open` to mint a short-lived WSS URL, then speak Slack's envelope framing over it (`hello`/`disconnect` control frames, envelope acks, ping/pong, and a reconnect that re-mints the URL). Without `slack-sdk` we would hand-roll all of that, plus `auth.test` for the identity assertion.

Added a comment in `pyproject.toml` saying so, and noting it is only needed when `AUTOMATION_STREAMS_ENABLED` is on.

pyproject.toml
VascoSch92: We have `httpx`, but it cannot do this — httpx has no WebSocket client at all, and Socket Mode is a WebSocket. `slack_sdk.socket_mode.aiohttp.SocketModeClient` is built on aiohttp.

The reason it is a *direct* dependency rather than left implicit: `slack-sdk` declares aiohttp only under its `optional` extra, so installing `slack-sdk` alone does not bring it. Using `slack-sdk[optional]` instead would also pull in boto3, SQLAlchemy, websockets and websocket-client, which we do not want.

Worth knowing it is not a new wheel in the image either way — `uv tree` shows aiohttp 3.13.4 already arriving via `cloud-sql-python-connector` and via litellm under `openhands-sdk`. But relying on someone else's transitive dep for an import we make is the fragile version. Comment added in `pyproject.toml`.

AGENTS.md
VascoSch92: Removed — the whole cutover paragraph is gone from `AGENTS.md`. That is deployment runbook material, not repo documentation; it stays in the PR description and the issue.

AGENTS.md
VascoSch92: Cut from 47 lines to 30: dropped the cutover paragraph entirely, merged the "off by default" and "self-hosted" bullets into one, and tightened the rest. Same seven points, less prose.

openhands/automation/config.py
```diff
@@ -402,6 +403,49 @@ def enabled(self) -> bool:
...
+    """Stream sources: long-lived inbound connections, supervised in-process.
+
+    Off by default, and a self-hosted capability for two independent reasons:
+    Slack does not allow Socket Mode apps in the public Marketplace, and
+    connection-scoped state does not fit a stateless autoscaled tier. Webhooks
+    remain the cloud path.
+
+    Environment variables (AUTOMATION_ prefix):
+        AUTOMATION_STREAMS_ENABLED: Master switch (default: false). With it
+            off the supervisor never starts and nothing else changes.
```
enyst: I think maybe defaulting to True makes a little more sense to me. 🤔 Why would people need to flip an env var? Slack is recommended in Automation examples in canvas, and socket mode is, from what I can tell, the most appropriate for it. But no biggie, just a passing thought!

VascoSch92: It makes sense actually... in the VM i had to switch to True... not great

VascoSch92: Done — flipped to `True`.

You are right that the flip was pure cost. `enabled` was already `streams_enabled and slack_apps`, so the empty-`AUTOMATION_SLACK_APPS` case starts nothing either way — the switch was a second gate in front of a gate that already held. Configuring the apps is now the whole switch, which is what makes flipping the default safe: a deployment that configures none is still byte-for-byte today's behaviour.

Kept the var as a kill switch (`AUTOMATION_STREAMS_ENABLED=false`) so the sockets can be held down without unsetting credentials — that is the one thing dropping it entirely would have lost.

Tests follow the same shape: the app-configuration test no longer sets the var at all (proving apps alone suffice), and the old "switch alone starts nothing" test is now "switch off kills configured apps". Docstrings in `config.py`/`streams/__init__.py`, the `AGENTS.md` bullet and the `pyproject.toml` comment all updated to say the same thing.

openhands/automation/config.py
```diff
@@ -402,6 +403,51 @@ def enabled(self) -> bool:
         return bool(self.git_sync_repo_url and self.git_sync_enabled)
 
 
+class SlackAppSettings(BaseModel):
```
malhotra5, line 406: Just to confirm, it seems like the intended behavior is to be able to support multiple streams but considering that configuration in done via env will env name conflicts be a problem?

openhands/automation/app.py
```diff
@@ -195,6 +196,19 @@ async def lifespan(app: FastAPI):
         app.state.git_sync_task = git_sync_task
         logger.info("Background git sync started")
 
+    # Stream sources: long-lived inbound connections (Slack Socket Mode),
+    # one supervised task each. Starts only once an app is configured.
+    streams_task = None
```
malhotra5, line 201: Noting that streams are meant to be configured on startup, as opposed to mounted later

malhotra5 (approved):
LGTM

Issue #105, opened 2026-05-08
Feature: Session-based conversation reuse for event-triggered automations


malhotra5:
## Summary

Add support for reusing conversations across multiple events within a "session". This enables use cases like:
- Slack threads: Multiple messages in the same thread should go to the same conversation
- GitHub PRs: Multiple comments/events on the same PR should share context
- Linear issues: Follow-up events on the same issue should maintain conversation history

## Problem Statement

Currently, each event creates a new `AutomationRun` → new sandbox → new conversation. There is no mechanism to route related events to the same conversation. This means:
- No context sharing between related events
- Each event starts fresh without knowledge of previous interactions
- Expensive: new sandbox for every event even if related

### High-Level Flow

1. **Automations configure a session key expression** - JMESPath to extract a session identifier from event payloads
2. **First event creates a session** - AutomationRun with session_key, sandbox started, session state stored
3. **Subsequent events match to existing session** - Events with same session_key are routed to existing sandbox
4. **Events are queued to agent-server** - New endpoint allows automation service to push events
5. **SDK polls for events** - User's SDK script decides when/how to process queued events
6. **Handle sandbox death gracefully** - Configurable behavior: queue, restart, or drop

### Why Event Queue (vs Direct Conversation Manipulation)

We considered directly calling agent-server conversation APIs (`POST /api/conversations/{id}/events`), but this approach is inflexible:
- **Custom SDK code**: Users may have custom tarballs with their own conversation logic, filters, multiple conversations
- **State uncertainty**: Even if sandbox is alive, the SDK script might have exited, conversation might be closed
- **No user control**: The automation service would be making assumptions about conversation lifecycle

Instead, we push events to a **queue that the SDK can consume**, giving users full control over processing.

### Schema Changes

```python
class SessionConfig(BaseModel):
    """Configure session-based conversation reuse."""
    
    key_expr: str = Field(
        ...,
        description="JMESPath expression to extract session key from event payload"
    )
    
    idle_timeout_seconds: int = Field(
        default=300,
        ge=30,
        le=3600,
        description="SDK exits if no events for this long (prevents infinite sandbox)"
    )
    
    session_timeout_seconds: int = Field(
        default=3600,
        ge=60,
        le=86400,
        description="Max total session lifetime"
    )
    
    on_sandbox_death: Literal["queue", "restart", "drop"] = Field(
        default="queue",
        description=(
            "Behavior when sandbox dies while events are pending:\n"
[...]

PR #385, opened 2026-08-25
feat: route events to an existing conversation via a derived conversation id

### LINKED ISSUE #362: Route events to an existing conversation via an external subject mapping

VascoSch92:
Closes #362. Phase 5 of #363. **Stacked on #384** — review that first; this branch contains its commits and its base is `vasco/slack-socket-streams`.

Every matched event creates a new run, and therefore a new conversation with no memory of what came before. On the OSS VM you can @-mention the bot, get an answer, reply in-thread, and be answered as though the thread had never happened.

This adds **no table**. Per @enyst on #362, the correspondence between an external subject and its conversation is a documented deterministic rule, not stored state. What does get stored is two nullable columns recording which run was about which subject, and when it stopped being reachable — because the *sandbox* holding a conversation can only be named by the server.

## Blocking dependency: confirmed

The issue asks for this to be established before the phase starts. It works today, on the pinned SDK (1.43.1):

`POST /api/conversations/{id}/events` with `{"role": "user", "content": [...], "run": true}` appends the message and then, in `EventService.send_message`, starts the agent loop **only if the execution status is not already RUNNING**. That is exactly "a new turn on an idle conversation", and it is also correct for a conversation that is still busy — the message is queued into a live loop rather than starting a second one. No agent-server change is needed.

## What this adds

**The rule** (`subjects.conversation_id_for`):

```
conversation_id = uuid5(NS, f"{org_id}/{automation_id}/{source}/{subject_key}")
```

`NS` is pinned permanently — changing it re-keys every live thread at once, which reads to users as every conversation losing its memory on deploy.

`automation_id` is in the key deliberately. A conversation belongs to the automation whose script created it — its prompt, its tools, its tarball. Two automations in one org both watching `#engineering` would otherwise share a conversation, and the second would send turns into one that cannot answer them. It also means editing an automation re-keys its threads, rather than pinning a thread to whatever agent the first event saw — and attaching with a *different* agent kind raises server-side (`_agent_kind_mismatch_message`), so re-keying is the behaviour we want, not a compromise.

The derived id is handed to the run that opens the conversation as `AUTOMATION_CONVERSATION_ID`. Both presets read it and pass it to `Conversation(...)`; a test fails if a preset stops doing so, since a preset that ignored it would silently give every event a fresh thread. A custom tarball that ignores it gets the same fresh conversation per event, which is the documented contract rather than a silent failure.

**`AutomationRun.subject_key` and `subject_released_at`** (migration `022`) — two nullable columns behind one partial index. This is not a mapping to a conversation; the conversation id is derived and never stored. It is how a later event finds the sandbox still holding that conversation. `subject_released_at` records the moment a run stopped being that subject's routing target, so the key survives as history instead of being wiped. The index covers only live subjects.

**`subjects.py`** — the rule, and nothing else. There is no built-in subject extractor for any source: a trigger names its subject with `subject_key_expr`, or it does not have one. The two worked examples from the issue live in the `EventTrigger` docstring and the field description, with tests pinning them to that description so the docs and the tested strings cannot drift.

Slack is `team/channel/thread_ts`, falling back to `ts`, because the mention that *opens* a thread carries no `thread_ts` and its own `ts` is what becomes the thread id when somebody replies — without the fallback the opening mention and the reply to it would be two subjects:

```
join('/', [team_id, event.channel, event.thread_ts || event.ts])
```

GitHub is `owner/repo#number`, reading `pull_request`, `issue` and a bare top-level `number` to the same key, since a review on PR 12 and an `issue_comment` on PR 12 are one thing to a person. The leading guard keeps a `push`, which has no number, from becoming the subject `<repo>#null` — which is a real answer: there is no conversation to continue for a branch.

**`EventTrigger.destination`** — `dispatch_run` (default, today's behaviour) or `continue_conversation`. Per **trigger**, not per automation (open question 4): a trigger is where the event shape is already known, and #118's "opt-in per route/trigger" is the principle being reused.

**`EventTrigger.subject_key_expr`** — the `key_expr` concept from #118, kept, and the only way to name a subject. Validated at creation with the same JMESPath compile check `filter` uses, because a typo would otherwise look exactly like the feature not being switched on. Non-scalar results are refused — a dict is not an identity, and `true` would collapse every event onto one subject. An over-long or blank key is refused too: the column is `String(500)`, so an oversized key would be a failed INSERT rolling back the whole delivery.

**`EventTrigger.turn_text_expr`** — renders the event as the turn's text. Without one the event is rendered as the message a human actually wrote, and only a shape nothing recognises travels as verbatim JSON. Dumping the payload unconditionally buried the one line somebody wrote under ~15 KB of webhook metadata, every turn.

**`EventTrigger.wake_agent`** — `false` appends the turn and leaves it unanswered, making the conversation an ordered, durable buffer the script drains on its own terms.

**Routing**, in `accept_event`, per matched automation:

```
event → derive subject_key → take the subject's lock
      → a run already owns it?
          not started yet →  fold the turn into its payload, create no run
          started         →  derive the conversation id, send a turn, create no run
      → otherwise          →  create a run, carrying the subject on it
```

**Sandbox retention** — a `continue_conversation` automation is forced `keep_alive` at creation and on update, and rejected if it asks for `keep_alive: false`. Settling it there means every cleanup path keeps checking `keep_alive` and nothing else, instead of each one growing a second condition. Deleting the sandbox would destroy the conversation the next event is meant to continue, which would make the whole feature a no-op in cloud mode. The runtime TTL reaper still collects it. Cancelling a run is the one path that deletes the sandbox anyway, and it releases the subject as it goes.

## What the derived id buys over a mapping table

Beyond dropping `external_conversations` (table, model, and 67-line migration) for two columns, it closes a hole the mapping had.

A stored mapping cannot name a conversation until the run that created it **completes** and posts the id back on `/runs/{id}/complete`. So an event arriving while the first run was still going had nothing to continue, and forked a second run against the same thread. A derived id is known the moment the subject is, so that event continues the live conversation instead — the agent server appends the message and runs it only if the loop is not already going.

## Open questions, answered

1. **Expiry policy.** Nothing to expire — there is no mapping to go stale. The *sandbox* TTL is the real expiry, and when it fires the next event simply starts a run.
2. **Missing or failed conversation.** Degrades to a run. A finished run that can no longer be reached is stamped `subject_released_at`, so the row keeps both the subject and the moment it stopped routing, and later events stop paying the timeout to rediscover the same reaped sandbox. A run that is still going is never released, because it may not have recorded its sandbox yet. A reaped sandbox is ordinary, not an error, so this is logged at INFO and never raises.
3. **Serialising concurrent events.** A `pg_advisory_xact_lock` keyed on the automation and subject, taken before the lookup, plus `SELECT ... FOR UPDATE` on the run row it finds. The advisory lock is the load-bearing one: a row lock only orders events that find a row, and the burst that matters is two events that both find nothing. Both are transaction-scoped, so the lock also covers the run the event goes on to create. SQLite runs single-process and skips it. `get_event_automations` orders by `Automation.id` so an event matching several automations locks their rows in the same order in every worker — without that, two events on one thread can lock the same rows in opposite orders and deadlock.
4. **Per trigger or per automation.** Per trigger, above.

## Known remaining fork

The dispatcher marks a run RUNNING and commits before it provisions anything, and `sandbox_id` is only persisted after the bash command starts. In cloud mode that window spans sandbox provisioning plus tarball upload. An event arriving inside it finds a run that has started, so the turn is not folded in, but no sandbox to post to either — and it creates a second run on the same subject.

Two events on a *brand-new* subject no longer fork: the advisory lock orders them and the second is folded into the first's payload before it is dispatched. An event arriving *mid-run*, once the sandbox is recorded, does not fork either. Only the provisioning window above does.

The fix is to persist `sandbox_id` as soon as the execution context is ready and to fold a turn into any non-terminal run that is not yet routable, rather than keying on `started_at`. Thanks @malhotra5 for spotting it.

## Not adopted from #118

`PendingSessionEvent`, any SDK-side polling loop inside the sandbox, session lifecycle states owned by the service, sandbox restart plus event requeue. None of it is here. This is deliberately the narrower direct-continuation model, not the full session architecture from #105.

## Acceptance criteria

- [x] Two `app_mention` events in the same Slack thread reach the same conversation, and the second carries the new event so the first is still in history.
- [x] Two events in *different* threads do not.
- [x] Automations that do not opt in behave exactly as today — no subject, no lookup, no turn.
- [x] A subject whose sandbox is gone degrades to creating a new run rather than erroring.
- [x] Concurrent events for one subject do not race into two conversations.

## A bug found and fixed by self-review

`continue_conversation` fetched the run with `session.get()`, which leaves the
`automation` relationship unloaded. Minting a cloud API key reads
`run.automation`, so that access was a lazy load — `MissingGreenlet` under an
async session. Because `send_conversation_turn` catches everything and returns
False, it did not raise: it **silently fell back to creating a run, making the
whole feature a no-op in cloud mode.**

It happened to work when called from `accept_event`, because
`get_event_automations` has already put the automation in the session's
identity map and SQLAlchemy's many-to-one loader resolves from there without
SQL. An accident of call order, not a guarantee. Fixed with an explicit
`selectinload`, plus a regression test that drives the real ORM and backend
path with stubs only at the HTTP boundary — it fails on `session.get()` and
passes on the eager load.

## Testing

80 tests across `tests/test_conversations.py` and `tests/test_conversation_turn.py`: both documented subject expressions and their edge cases, `subject_key_expr` resolution and rejection, trigger validation, every routing branch, per-automation scoping (two automations on one subject get two distinct derived ids), coalescing into a queued run, mid-run continuation, a real two-transaction concurrency test against Postgres, the completion callback holding the sandbox, cancellation releasing the subject, and the exact request made to the agent server.

`1659 passed, 7 skipped` on the full unit suite. `pre-commit` (ruff format, ruff lint, pycodestyle, pyright) clean.

HUMAN: An earlier revision of this branch was run end to end against a real HTTP agent server, not just the test suite. That run predates the switch from a mapping table to a derived conversation id, so it does not cover the current routing rule and is being repeated before this leaves draft.

**Still to redo before this leaves draft:** that end-to-end run was done against the mapping design and does not cover the derived id, so it needs repeating — specifically that the first run really does open the conversation under `AUTOMATION_CONVERSATION_ID`, which is the assumption the whole routing rule rests on.

## Out of scope

A Canvas UI for the destination field, and subject extractors for `jira_dc` and `bitbucket_data_center` (both reachable today via `subject_key_expr`). The event payload is handed over verbatim when no `turn_text_expr` is set and nothing in it is recognisable, which is the shape the automation's own script was given for the first turn.


openhands/automation/conversations.py
```diff
@@ -0,0 +1,159 @@
...
+    event_payload: dict[str, Any] | None,
+) -> str | None:
+    """Deliver this event as another turn on the subject's conversation.
+
+    Returns the conversation id when the turn landed, None when the caller
+    should create a run: nothing has run for this subject yet, or its sandbox
+    is gone. Holds the run's row lock until the caller commits.
+
+    The id is known before the first run finishes, so an event arriving
+    mid-run continues that conversation instead of racing a second run.
```
enyst, line 259: Is this necessary? A user message sent to a mid-turn conversation in the SDK is naturally added to its history. Does Automation need to code it again, maybe… 🤔

openhands/automation/ingest.py
```diff
@@ -73,6 +81,11 @@ async def accept_event(
     `event.provider_event_id` when the transport supplies one. Commits before
     returning.
 
+    A matched automation whose trigger sets `destination` to
+    `continue_conversation` first tries to deliver the event as another turn on
+    its subject's conversation, creating no run. Anything else -- no subject,
```
enyst: Why does it create no run, when it continues?

Isn’t this like:
- automation run: a triggered event sends something to the SDK
- conversation: a SDK concept; recognized if existing, newly created if not pre-existing

enyst: I think, though I could be wrong, that the same goes for the enterprise/cloud app-server: 
- if I use an OH Cloud API key to send a message to a conversation ID, I think it should do the necessary (provision a sandbox or find an older preexisting sandbox) and send it the event… 🤔

The reason I’m saying this is that it’s a bit unclear to me if or why we seem to re-do processing that the app-server must do anyway, but maybe I’m missing something

malhotra5 (changes requested):
Some immediate concerns I have

1. when sandbox has been paused it seems we will start a new conversation run (which can be confusing from UX perspective)

2. we're routing based on a deterministically generated conversation ID, but the run initially generates a random conversation UUID 

3. currently seems less flexible than https://github.com/OpenHands/automation/issues/105



As I understand it, the current idea is that if an existing subject conversation is reachable, append another user turn to it; otherwise create a new run.

IMHO the separation should be session lifecycle belongs to the automation service, and follow-up event processing belongs to the SDK script. This would provide maximum flexibility and control as opposed to always routing incoming events into a running agent conversation (e.g we may want to perform deterministic processing for new incoming events and selectively pass some events to the agent conversation which the sdk script should control)

openhands/automation/presets/plugin/sdk_main.py
```diff
@@ -532,6 +533,15 @@ def event_callback(event) -> None:
     }
     if automation_user_id and _conversation_supports_user_id():
         conversation_kwargs["user_id"] = automation_user_id
+    # Set by the service for a `continue_conversation` run: the conversation
```
VascoSch92: trimmer the comment

openhands/automation/presets/prompt/sdk_main.py
```diff
@@ -485,6 +486,15 @@ def event_callback(event) -> None:
     }
     if automation_user_id and _conversation_supports_user_id():
         conversation_kwargs["user_id"] = automation_user_id
+    # Set by the service for a `continue_conversation` run: the conversation
```
VascoSch92: trimmer the comment

openhands/automation/utils/conversation_turn.py
```diff
@@ -0,0 +1,144 @@
...
+
+# Short: the caller holds the subject's row lock for the length of this call.
+TURN_TIMEOUT_SECONDS = 15.0
+
+# An idle sandbox is paused, not deleted, so the conversation is still there.
+# Waiting for it costs lock time; losing the thread's memory costs the user
+# more. Past the budget we give up and the caller starts a run, as before.
+RESUME_WAIT_SECONDS = 20.0
+RESUME_POLL_SECONDS = 2.0
+
```
VascoSch92: use Final[int] here

openhands/automation/utils/webhook.py
```diff
@@ -153,8 +153,11 @@ async def get_event_automations(
             Automation.trigger.op("->>")("source") == literal(source),
         )
 
+    # Ordered so an event matching several automations locks their subject
```
VascoSch92: trimmer

openhands/automation/conversations.py
```diff
@@ -0,0 +1,171 @@
...
+from openhands.automation.subjects import EventSubject, conversation_id_for
+from openhands.automation.utils.conversation_turn import (
+    compose_turn,
+    send_conversation_turn,
+)
+
+
+logger = logging.getLogger("automation.conversations")
+
+CONTINUE_CONVERSATION = "continue_conversation"
```
VascoSch92: use Final[str]

openhands/automation/dispatcher.py
```diff
@@ -326,6 +327,21 @@ async def _fail(
     env_vars["AUTOMATION_EVENT_PAYLOAD"] = json.dumps(
         _build_event_payload(automation, run)
     )
+    # A subject-owning run has to create its conversation under the very id
```
VascoSch92: Trimmer the comments

openhands/automation/models.py
```diff
@@ -200,6 +200,11 @@ class AutomationRun(Base):
     # The sandbox ID used for execution (for status verification)
     sandbox_id: Mapped[str | None] = mapped_column(String(255), nullable=True)
 
+    # The external subject this run is about (a Slack thread, a PR). Not a
```
VascoSch92: trimmer the comments

openhands/automation/conversations.py
VascoSch92: the first turn is written by the automation's own script, the continuation by a fixed template here, because the script doesn't run on a continue. The payload goes over verbatim so nothing is lost, but the framing differs between turn 1 and turn 2. If that bothers you, I'd rather add a turn template on the trigger than move rendering into the script.

openhands/automation/ingest.py
VascoSch92: On the missing run: deliberate, but you're right that it costs something.

I'm happy to write a run row for a continuation instead, marked as such and skipping dispatch. It keeps your model (run = we sent something to the SDK), costs one INSERT, and still creates no sandbox. 

On the app-server: the endpoint you're describing does exist and it does resolve the sandbox itself. It just doesn't know about our conversations. We create a bare sandbox via POST /api/v1/sandboxes and the SDK script creates the conversation against the agent server inside it, so there's no AppConversation row and that route 404s for everything we own.

It also doesn't provision or resume: a paused sandbox gets a 409 telling you to call /sandboxes/{id}/resume first.

malhotra5:
Thanks for the detailed implementation and write-up. I reviewed this against the earlier session-based proposal in #105, and I think the two are aiming at the same product problem but make a fairly different architectural tradeoff.

High-level read:

- #105 proposes a session-and-event-queue architecture: `session.key_expr`, explicit session lifecycle, pending event storage, SDK polling/acking, idle/session timeouts, and sandbox-death policies like queue/restart/drop.
- This PR implements a narrower direct-continuation model: `destination: continue_conversation`, `subject_key_expr`, provider-derived subjects, deterministic `uuid5` conversation IDs, and one nullable `automation_runs.subject_key` column used to find the sandbox/run that should receive follow-up events.

The concepts are closely related. `subject_key_expr` is effectively the identity/grouping half of `session.key_expr`: both answer “what external thing are these events about?” — Slack thread, GitHub PR, Linear issue, Jira ticket, etc. The semantic drift is that `session.key_expr` belongs to a first-class managed session, while `subject_key` is only a routing key for attempting to continue an existing conversation. That makes `subject_key` an accurate name for this PR, but it does mean this is not the full session architecture from #105.

Where this PR is strong:

- Much smaller implementation surface: no new session table, no pending event queue, no SDK/agent-server queue API.
- Uses existing agent-server behavior via `POST /api/conversations/{id}/events` with `run: true`.
- Keeps current behavior as the default via `dispatch_run`.
- Per-trigger opt-in is clean and practical.
- Deterministic conversation IDs avoid the obvious mapping-table race where a second event arrives before the first run has completed and reported its conversation ID.
- Built-in Slack/GitHub subject derivation looks sensible.

What this PR does not cover from #105:

- No `AutomationSession` lifecycle/state.
- No `PendingSessionEvent` queue.
- No SDK-side `get_pending_events()` / `ack_event()` flow.
- No idle/session timeout knobs.
- No configurable sandbox-death policy (`queue` / `restart` / `drop`).
- No guaranteed preservation of events that arrive before a first sandbox is ready or while a target sandbox is gone.
- Less control for custom tarballs, because the service formats and injects the follow-up turn directly.

One specific thing I would like to double-check before relying on this approach: the routing conversation ID is derived by the automation service from `org_id / automation_id / source / subject_key`; it is not read from the agent server or from `AutomationRun.conversation_id`. That means the first run must create/use that same deterministic conversation ID. While inspecting the automation repo changes, I did not see the preset `Conversation(...)` construction being passed the derived ID for the initial run. If the first run creates a random conversation ID, follow-up routing to the derived ID could miss and fall back to a new run. Maybe this is handled by SDK behavior or another layer, but I think it is worth validating explicitly with the E2E test the PR body mentions still needs to be redone.

Preference/request: if possible, we would like to prioritize the architecture in #105 — the session + queue model — because it gives us the more general and reliable contract for event-triggered automations: explicit lifecycle, event buffering, SDK-owned processing, and configurable behavior when sandboxes die. This PR may still be a useful lightweight v1, but I would prefer not to treat it as equivalent to the #105 design unless we are intentionally deciding to drop the queue/session semantics.

AI disclosure: This comment was created by an AI agent (OpenHands) on behalf of the user.

malhotra5:
Noting for posterity, 

current PR seems reasonable to begin with and sandbox lifecycle management can be a next step

we may have a race condition in a multi-pod envs where multiple events in a quick burst and can cause duplicate runs

```
first event
→ create PENDING AutomationRun(subject_key=...)

second event arrives before first run has started
→ no started run found
→ create another AutomationRun(subject_key=...)
```


e.g scenario where two people trigger slack resolver back to back - we can use the DB as a queue for incoming events and poll of it for session reuse

malhotra5:
NIT: thoughts on organizing session reuse related things as a nested object instead of on root level?

e.g 
```
{
  "type": "event",
  "source": "github",
  "on": "issue_comment.created",
  "session": {
    "key_expr": "join('', [repository.full_name, '#', to_string(issue.number)])",
    "desintation": "continue_conversation",
  }
}
```

In the future we may expand the session object to include sandbox management configs, etc

openhands/automation/utils/conversation_turn.py
```diff
@@ -0,0 +1,278 @@
...
+TURN_TIMEOUT_SECONDS: Final[int] = 15
+
+# An idle sandbox is paused, not deleted, so the conversation is still there.
+# Waiting for it costs lock time; losing the thread's memory costs the user
+# more. Past the budget we give up and the caller starts a run, as before.
+RESUME_WAIT_SECONDS: Final[int] = 20
+RESUME_POLL_SECONDS: Final[int] = 2
+
+
+# Where a human-authored message sits in a webhook payload, most specific
```
VascoSch92: Trimmer the comment here

openhands/automation/conversations.py
```diff
@@ -0,0 +1,312 @@
...
+        return None
+    return key
+
+
+async def _take_subject_lock(
+    session: AsyncSession,
+    automation_id: uuid.UUID,
+    subject_key: str,
+) -> None:
+    """Serialise every event for one subject, including those finding no run.
```
VascoSch92, line 157: can you trimmer the comment?

openhands/automation/conversations.py
```diff
@@ -0,0 +1,312 @@
...
+        )
+    )
+
+
+async def _lock_subject_run(
+    session: AsyncSession,
+    automation_id: uuid.UUID,
+    subject_key: str,
+) -> AutomationRun | None:
+    """The most recent run holding this subject, started or not, locked.
```
VascoSch92, line 187: Trimmer the comment

openhands/automation/conversations.py
```diff
@@ -0,0 +1,312 @@
...
+
+    run = await _lock_subject_run(session, automation_id, subject_key)
+    if run is None:
+        return ContinueResult()
+
+    conversation_id = conversation_id_for(org_id, automation_id, source, subject_key)
+    turn = compose_turn(source, event_key, event_payload, override=turn_text)
+
+    if run.started_at is None:
+        # Queued, not gone: there is no agent server yet, but this run already
```
VascoSch92: trimmer the comment

openhands/automation/conversations.py
```diff
@@ -0,0 +1,312 @@
...
+            run.id,
+            conversation_id,
+        )
+        return ContinueResult(conversation_id=conversation_id, coalesced=True)
+
+    delivered = await send_conversation_turn(
+        run, conversation_id, turn, wake_agent=wake_agent
+    )
+    if not delivered:
+        # The conversation could not be reached. Only forget a finished run --
```
VascoSch92: trimmer the comment

openhands/automation/schemas.py
```diff
@@ -160,16 +188,68 @@ class EventTrigger(BaseModel):
         ),
     )
 
-    @field_validator("filter")
+    destination: Literal["dispatch_run", "continue_conversation"] = Field(
```
VascoSch92, line 227: you can trimmer the description of every field here

openhands/automation/subjects.py
```diff
@@ -0,0 +1,81 @@
...
+without a cycle.
+"""
+
+import uuid
+from dataclasses import dataclass
+from typing import Any
+
+
+# Pinned permanently: changing it makes every live thread lose its memory.
+CONVERSATION_NAMESPACE = uuid.UUID("d7f3a2b1-5c48-4e9a-9b6d-2f1e8c3a7d40")
```
VascoSch92: type

openhands/automation/subjects.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@dataclass(frozen=True, slots=True)
+class EventSubject:
+    """The external thing an event is about."""
+
+    key: str
+
```
VascoSch92: Do we need that? Can we just not do a type instead of this class?

openhands/automation/providers.py
```diff
@@ -292,18 +293,20 @@ def _register_builtin_providers() -> None:
...
-    # deliveries; the other two are recorded without being deduplicated.
-    for source, parse, event_id_header in (
-        ("bitbucket_data_center", parse_bitbucket_data_center_event, None),
-        ("github", parse_github_event_auto, "X-GitHub-Delivery"),
-        ("jira_dc", parse_jira_dc_event, None),
+    # deliveries; the other two are recorded without being deduplicated. Only
+    # GitHub has a subject extractor; the others use `subject_key_expr`.
+    for source, parse, event_id_header, subject in (
+        ("bitbucket_data_center", parse_bitbucket_data_center_event, None, None),
+        ("github", parse_github_event_auto, "X-GitHub-Delivery", github_subject),
```
malhotra5: i thought a subject is defined on a per-automation basis

are we assigning this as the default subject for the provider?

malhotra5: so just to confirm the idea is that

1. event comes in
2. filter automations that match incoming event
3. apply subject key expr derived from the qualified automation
4. reroute to existing runs/create new run based on subject key expr result

This imho seems cleaner, the expression is held on the automation definition level

openhands/automation/router.py
```diff
@@ -654,7 +654,11 @@ async def complete_run(
 
     # Clean up immediately when this automation owns explicit cleanup. Once
     # post-run callbacks exist, this path should run them before deleting.
-    if run.sandbox_id and automation.keep_alive is not True:
+    #
+    # A run owning a subject is treated like keep_alive: deleting its sandbox
+    # would destroy the conversation the next event continues. The runtime TTL
+    # reaper still collects it.
+    if run.sandbox_id and automation.keep_alive is not True and not run.subject_key:
```
malhotra5: I'm worried we're complicating the conditionals here


What if we enforce that `keep_alive=False` and `subject_key=xxx` cannot be true at the same time on the automation creation endpoint? That way we can keep the conditional here the same

openhands/automation/router.py
```diff
@@ -813,8 +817,15 @@ async def cancel_run(
...
 
-    # Clean up sandbox for runs that were RUNNING
+    # Clean up sandbox for runs that were RUNNING. Cancelling is explicit, so
+    # unlike `complete_run` the sandbox goes even when the run owns a subject
+    # -- but the subject goes with it, or the next event would pick this run
+    # and pay a lookup for a sandbox we just deleted.
     if run.sandbox_id:
+        if run.subject_key:
+            run.subject_key = None
+            await session.commit()
```
malhotra5: I'd like to keep the subject key value as a historical record, could make for easier debugging in the future

Is it possible to find another way to skip runs where the sandbox has been deleted as opposed to erasing the data?

malhotra5:
One remaining race seems possible outside the pure `PENDING` burst case:

```text
run marked RUNNING / started_at set
but sandbox_id or conversation route not recorded/reachable yet
→ follow-up arrives
→ started_at is not None, so it does not coalesce
→ send_conversation_turn cannot resolve the agent server
→ falls back to creating a new run
```

So the PR appears to handle follow-ups while `PENDING` and while fully routable, but the RUNNING/provisioning gap can still fork. Suggest treating a non-terminal subject-owning run that is not yet routable as coalescable/provisioning, or modeling this via session state/pending events.

AI disclosure: This comment was created by an AI agent (OpenHands) on behalf of the user.

openhands/automation/providers.py
VascoSch92: Yes, your four steps are exactly the flow. 

The extractor on the Provider is only a built-in default so the common GitHub case works with no expression written. 

The automation's own `subject_key_expr` always wins when it is set, and it is evaluated
against the raw payload after the automation matches.

So the expression does live at the automation definition level, as you prefer.

openhands/automation/router.py
VascoSch92: I think you are right. 

I changed :-)

openhands/automation/router.py
VascoSch92: Yes.

Added `subject_released_at`, and the lookup now filters on it being null instead. Cancel stamps the timestamp rather than wiping the key, and the failed-turn path does the same, so the row keeps both the subject and
the moment it stopped routing.

VascoSch92:
> One remaining race seems possible outside the pure `PENDING` burst case:
> 
> ```
> run marked RUNNING / started_at set
> but sandbox_id or conversation route not recorded/reachable yet
> → follow-up arrives
> → started_at is not None, so it does not coalesce
> → send_conversation_turn cannot resolve the agent server
> → falls back to creating a new run
> ```
> 
> So the PR appears to handle follow-ups while `PENDING` and while fully routable, but the RUNNING/provisioning gap can still fork. Suggest treating a non-terminal subject-owning run that is not yet routable as coalescable/provisioning, or modeling this via session state/pending events.
> 
> AI disclosure: This comment was created by an AI agent (OpenHands) on behalf of the user.

I fixed the second one: when the sandbox is up but the script hasn't opened the
conversation yet, the append retries for 10s instead of 404ing straight into a new run.

The first one is still there. We hold the run's row lock for the whole attempt, and the dispatcher needs that lock to write sandbox_id, so waiting inside it just blocks the write we're waiting for. 

Either we record sandbox_id as soon as the sandbox exists, which shrinks the window to almost nothing,
or we park the event, which is your queue.

Could we do in a follow-up PR?

openhands/automation/subjects.py
```diff
@@ -0,0 +1,79 @@
...
+    if not isinstance(event, dict):
+        return None
+    channel = event.get("channel")
+    thread = event.get("thread_ts") or event.get("ts")
+    if not team or not channel or not thread:
+        return None
+    return f"{team}/{channel}/{thread}"
+
+
+def github_subject(payload: dict[str, Any]) -> EventSubject | None:
```
malhotra5: Can we remove this? It seems that subject keys are on the automation level what is the reason for having this on the provider level?

openhands/automation/subjects.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    return str(
+        uuid.uuid5(
+            CONVERSATION_NAMESPACE,
+            f"{org_id}/{automation_id}/{source}/{subject_key}",
+        )
+    )
+
+
+def slack_subject(envelope: dict[str, Any]) -> EventSubject | None:
```
malhotra5: ditto here - do we need this on the provider level?

openhands/automation/utils/conversation_turn.py
```diff
@@ -0,0 +1,296 @@
...
+    event_key: str,
+    payload: dict[str, Any],
+) -> str | None:
+    """The event as the message a human wrote, or None if none is recognisable."""
+    body, body_path = _first_hit(payload, _BODY_PATHS)
+    if not isinstance(body, str):
+        return None
+
+    author = _first(payload, _AUTHOR_PATHS)
+    repo = _first(payload, ("repository.full_name",))
```
malhotra5, line 132: why are we hard coding this message renderer based on repo full name? 

ideally things are all encoded as configurable jmes expression paramaters to keep the abstraction as flexible as possible

openhands/automation/subjects.py
VascoSch92: dropped :-)

plain expression

```
join('/', [team_id, event.channel, event.thread_ts || event.ts])
```

openhands/automation/utils/conversation_turn.py
VascoSch92: That's `turn_text_expr`, already exists, it just wasn't in the worked examples, which is on me. Added.

```json
{"source": "github", "on": "issue_comment.created",
 "destination": "continue_conversation",
 "turn_text_expr": "comment.body"}
```

That function is only the fallback for triggers that set nothing, and `repository.full_name` is one probe in a list rather than a requirement. When it misses you get "@someone sent a new `slack` message (`app_mention`)" and the body still comes through fine. Without it the default would be a 15 KB JSON dump of the whole webhook, which buries the one line a human actually wrote.

malhotra5 (approved):
LGTM

openhands/automation/utils/conversation_turn.py
malhotra5: Sounds good, a future PR should maybe place a limit on the payload length and reject accordingly; users could run into this problem for custom webhooks

VascoSch92:
@all-hands-bot please solve the conflicts

PR #388, opened 2026-08-26
feat: report live run phases for dashboard visibility

hieptl:
## Why

Users cannot tell what an automation run is doing while it is active — the API exposes only a coarse run status until the run terminates, so long or multi-step automations (e.g. a code-review automation that clones a repo, examines a diff, and posts comments) feel opaque: there is no way to see whether a run is progressing, stuck, or nearly done. Issue OpenHands/OpenHands#16572 asks for live, user-facing phase updates during a run.

There was no mid-run write channel at all: the only sandbox→service writes were the terminal `/complete` and `/cancel` endpoints, and the preset scripts printed their phase banners to stdout only.

## Summary

- Add `AutomationRun.current_phase` (nullable, 200 chars; migration `021`, revising `020`) plus a sandbox-callable `POST /v1/runs/{run_id}/phase` endpoint that reuses the completion callback's auth/ownership pattern, only accepts writes while the run is PENDING/RUNNING (409 once terminal), normalizes the message server-side, and exposes the field on `AutomationRunResponse`.
- Instrument the run pipeline end to end: the dispatcher injects `AUTOMATION_PHASE_URL` and writes "Preparing environment" / "Starting automation" infra phases; the preset `sdk_main.py` scripts report standard setup phases and then stream redacted, throttled LLM-authored per-action summaries as live phases; both `setup.sh` scripts report "Installing dependencies". All reporting is best-effort — a failed phase POST can never fail a run.
- Document the contract for custom automations in `docs/run-phase-reporting.md` (env vars, curl/Python examples, response codes, safety guidance: no secrets, payloads, or stack traces in phase text).

## Issue Number

Fixes OpenHands/OpenHands#16572 (backend half; the dashboard rendering lands in the GUI repository).

## How to Test

1. Run the suite: `make test` (or `uv run python -m pytest tests/ -q`) — see the new `TestReportRunPhase` and dispatcher phase tests.
2. Manual API check against a local service:
   - Dispatch a run, then while it is RUNNING:
     `curl -X POST "$BASE/v1/runs/<run_id>/phase" -H "Authorization: Bearer <key>" -H "Content-Type: application/json" -d '{"phase": "Examining the diff"}'`
   - `GET /v1/{automation_id}/runs` now returns `current_phase` on the run.
   - Repeat the POST after the run completes → `409`.
[...]

PR #389, opened 2026-08-26
feat: add editable automation descriptions

### LINKED ISSUE #16925: [Feature]: Give automations an editable description, shown in the UI
### Problem or Motivation

An automation carries a name and nothing else that describes it. The list and detail surfaces have nowhere to say what an automation actually *does*, so the card falls back to rendering the raw prompt as its subtitle — cards read as instructions addressed to an agent rather than a summary addressed to a person. Once a workspace holds more than a handful of automations, the list stops being scannable.

Renaming is already possible: the Edit dialog renders a Name input and PATCHes `/v1/{id}`, and the service has always accepted `name`. Description simply does not exist — there is no column on `automations`, no field on any request or response schema, and no property on the frontend `Automation` type.

### Desired Behavior

As a user, I can give any automation a short description, edit it later from the same dialog I rename it in, and read it wherever the automation is listed.

- The Edit dialog offers a Description field alongside Name and Prompt.
- The automation card's subtitle shows the description, falling back to the prompt when there is none, so undescribed automations look exactly as they do today.
- The list row and the detail page header both show it.
- An automation launched from a template arrives with the catalog blurb already filled in, rather than empty.
- The description survives export/import and git-sync round-trips.

### Scope

Three repos, in this order — each is one PR, and the last is blocked on the first two being released.

**1. `OpenHands/automation` — persist and expose the field**

- [ ] Migration `020_add_automation_description.py`: `description` (`Text`, nullable) on `automations` (latest revision on main is `019`)
- [ ] `description` on the `Automation` model, on `CreateAutomationRequest` / `UpdateAutomationRequest` (`max_length=2000`), and on `AutomationResponse`
- [ ] `create_automation` passes it through — the PATCH handler needs no change, it setattrs everything in `body.model_dump(exclude_unset=True)`
- [ ] Same field on `CreatePromptAutomationRequest` / `CreatePluginAutomationRequest` and both `Automation(...)` constructions in `preset_router.py`, so template-created automations can carry one
- [ ] `"automationDescription"` added to `_STATIC_FEATURES` in `capabilities_router.py`
- [ ] git-sync: serialize into `automation.yaml` from `_automation_yaml_fields` **only when non-null**, and read it back in `_validate_and_resolve_fields`
- [ ] Release to PyPI

**2. `OpenHands/extensions` — declare the attribute**

- [ ] `"description"` added to `AutomationAttributeName` in `automations/index.d.ts`
- [ ] `automations/interface.schema.json`: `"description": { "$ref": "#/$defs/attribute" }` under `attributes.properties` (the object sets `additionalProperties: false`)
- [ ] `automations/interface.json`: `{ "type": "textarea", "label": "Description", "required": false }`
- [ ] Release

**3. `OpenHands/OpenHands` — edit and display**

- [ ] Bump `versions.automation` in `config/defaults.json` and `@openhands/extensions` in `package.json`
- [ ] `description?: string | null` on the `Automation` type; `"description"` in `AutomationAttributeName` and in `HOST_ATTRIBUTES` (`ATTRIBUTE_NAMES` derives from it)
- [ ] Description textarea in `edit-automation-modal.tsx`, gated on both the manifest attribute and the `automationDescription` capability
- [ ] Card subtitle, list row, and detail header render it
- [ ] `buildCreatePayload` / `buildBundlePayload` in `src/manifests/automation-setup.ts` seed it from `entry.description`
- [ ] Export/import carry it (`serializeAutomation`, `parseAutomationFile`, `buildCreateAutomationRequest`)

### Acceptance Criteria

- [ ] A description set in the Edit dialog persists across a reload and appears on the card, the list row, and the detail page.
- [ ] Clearing the description sends `null` and the card falls back to showing the prompt, unchanged from today's rendering.
- [ ] An automation with no description renders exactly as it does today — no empty line, no layout shift.
- [ ] Renaming still works from the same dialog, and a rename sent together with a description edit succeeds as one PATCH.
- [ ] Against an automation service that predates the field, the Description input does not render and no request carries `description` — editing name, prompt, model, schedule, and timeout all still work.
- [ ] An automation launched from a template arrives with the catalog description already set.
- [ ] Exporting an automation with a description and re-importing it preserves the description; the export file version stays `1` and older files without the field still import.
- [ ] A git-synced automation with no description produces no change to its `automation.yaml` — enabling this feature does not re-export every already-synced automation.
- [ ] Editing `description` in a synced `automation.yaml` updates the automation on the next sync cycle.
- [ ] A screenshot of the Edit dialog and of the automations list is attached to the PR.

### Alternatives Considered

**Reuse the prompt as the description.** What the cards do today, and the reason they read poorly — a prompt is written for an agent and is often several hundred characters.

**Skip the extensions release and make Description a host-owned field.** Every input in the Edit dialog resolves through `getAttributeSpec()`, and the attribute set is published by `@openhands/extensions` in `automations/interface.json`; an attribute the manifest does not declare does not render. Rendering it outside that lookup would cut the change from three repos to two, but it breaks the rule the whole automations surface is built on — the host holds no automation-specific copy or field list. Rejected in favour of a three-file change plus a version bump.

**Store the description in `preset_metadata`.** Avoids a migration, but the field would exist only for preset-created automations, would not be settable through the normal update path, and would not round-trip through git-sync.

### Additional Context

Two traps worth knowing before starting:

- `UpdateAutomationRequest` sets `extra="forbid"`, so a PATCH carrying `description` to an older automation service returns 422 and fails the *entire* request — including a rename sent alongside it. This is why the UI gates the field on the `automationDescription` capability rather than on the manifest alone: the manifest ships with the app, but the service version is whatever the deployment happens to run.
- The git-sync exporter hashes the serialized file tree. Emitting `description: null` for automations that have none changes every hash on the first cycle and pushes one no-op commit per automation, so the key must be omitted when null — the same reasoning as the existing `tarball_executables` handling in `_automation_yaml_fields`.

Editing is currently gated to local backends (`canEdit = active.backend.kind === "local"`, in both `src/routes/automation-detail.tsx` and `src/routes/automations-list.tsx`), so this ships local-only. Lifting that gate for cloud is a separate change — it would expose every editable attribute at once, not just this one.

<sub>References verified against `OpenHands/OpenHands@59981caf7`, `OpenHands/automation@7e9b89a`, and `OpenHands/extensions@806dfe5`.</sub>

santhiprakash:
## Problem

An automation carries only a name — there is no field anywhere that describes what it *does*, so list/detail surfaces fall back to rendering the raw prompt as the subtitle and cards read as instructions addressed to an agent rather than a summary addressed to a person.

Implements **part 1** (this repo) of OpenHands/OpenHands#16925: persist and expose an editable `description`.

## Changes

- **Migration `022_add_automation_description`** — nullable `Text` column `description` on `automations` (cross-DB: generic `sa.Text`, upgrade + downgrade verified on SQLite). Renumbered to `022` after `main` merged `021_add_run_current_phase` (#388), so the chain stays single-head.
- **`models.py`** — `Automation.description: Mapped[str | None]`.
- **`schemas.py`** — `description: str | None` (`max_length=2000`) on `CreateAutomationRequest` and `UpdateAutomationRequest`; exposed on `AutomationResponse`. The PATCH handler needs no change (it already setattrs everything in `body.model_dump(exclude_unset=True)`, so an explicit `null` clears the field).
- **`router.py`** — `create_automation` passes `body.description` through.
- **`preset_router.py`** — same field on `CreatePromptAutomationRequest` / `CreatePluginAutomationRequest`, and both `Automation(...)` constructions, so template/preset-created automations can carry one.
- **`capabilities_router.py`** — `"automationDescription"` added to `_STATIC_FEATURES`, so the UI can gate the field on service version (an older service with `extra="forbid"` would 422 a PATCH carrying `description`).
- **`git_sync/serializer.py`** — `description` serialized into `automation.yaml` **only when non-null**, matching the existing `tarball_executables` reasoning: emitting `description: null` for automations that have none would rewrite every synced file (and its content hash) on the first cycle, pushing one no-op commit per already-synced automation.
- **`git_sync/loop.py`** — `_validate_and_resolve_fields` reads `description` back, so a git-side edit imports on the next sync cycle (both create and update paths).

Parts 2 (`OpenHands/extensions` manifest attribute) and 3 (`OpenHands/OpenHands` UI) are separate PRs per the issue's sequencing; this PR is self-contained and backwards-compatible.

## Verification

- `uv run pytest tests/ -q --ignore=tests/integration` → **1462 passed, 0 failed** (includes new tests below).
- Rebased onto `main` @ `f1b3244` (release 1.9.0; includes #393 SDK 1.44.0 bump and #388's `021_add_run_current_phase`): full suite → **1486 passed, 0 failed** on SDK 1.44.0; `alembic` history verified single-head `… → 020 → 021 → 022` with `current_phase` on `automation_runs` and `description` on `automations`, plus `downgrade -1` → `021` and re-`upgrade head` → `022` on a scratch SQLite DB; pre-commit (ruff format/lint, pycodestyle, pyright) all pass.
- `alembic upgrade head` then `alembic downgrade -1` against a scratch SQLite DB — column added/removed cleanly.
- `uv run pre-commit run --files <all changed files>` — ruff format, ruff lint, pycodestyle, pyright all pass.

New regression tests:

- `tests/test_router.py` — create with/without description (persisted + echoed, absent → `null`); 2001-char description rejected (422); PATCH sets and explicitly-null clears.
- `tests/test_git_sync_serializer.py` — `description` key omitted from `automation.yaml` when null; serialized when set; round-trips through `deserialize_automation`.
- `tests/test_git_sync.py` — full sync-cycle round trip: description lands in `automation.yaml`, and a git-side edit to it imports back into the DB on the next cycle.
- `tests/test_capabilities_router.py` — `automationDescription` advertised.

## Notes

- The only pre-existing test touched is the `AutomationResponse` constructor fixture in `tests/test_schemas.py` (added `description=None` to the defaults dict) — additive, no assertions changed.
- Fixes OpenHands/OpenHands#16925 (part 1 of 3).

HUMAN: no human review was performed on this change; authored by Santhi Prakash with AI assistance and verified as described above.

### DIFF
--- a/migrations/versions/022_add_automation_description.py
+++ b/migrations/versions/022_add_automation_description.py
@@ -1,0 +1,28 @@
+"""Add automations.description: optional human-facing summary.
+
+Revision ID: 022
+Revises: 021
+Create Date: 2026-08-26
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+
+revision: str = "022"
+down_revision: str = "021"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "automations",
+        sa.Column("description", sa.Text(), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("automations", "description")
--- a/openhands/automation/capabilities_router.py
+++ b/openhands/automation/capabilities_router.py
@@ -81,6 +81,10 @@ _STATIC_FEATURES = (
     "presetPlugin",
     "presetPrompt",
     "repoClone",
+    # Automations carry an editable human-facing description. The UI gates
+    # the field on this rather than the manifest alone: a PATCH carrying
+    # `description` to an older service 422s the whole request.
+    "automationDescription",
 )
 
 # Tags Pydantic inserts into an error location for the trigger union.
--- a/openhands/automation/git_sync/loop.py
+++ b/openhands/automation/git_sync/loop.py
@@ -488,6 +488,7 @@ async def _validate_and_resolve_fields(
         "enabled": True if fields.get("enabled") is None else bool(fields["enabled"]),
         "prompt": fields.get("prompt"),
         "preset_metadata": fields.get("preset_metadata"),
+        "description": fields.get("description"),
         "tarball_path": tarball_path,
     }
 
--- a/openhands/automation/git_sync/serializer.py
+++ b/openhands/automation/git_sync/serializer.py
@@ -151,6 +151,11 @@ def _automation_yaml_fields(
     # already-synced automation.yaml just to add an empty list.
     if tarball_executables:
         fields["tarball_executables"] = tarball_executables
+    # Same reasoning: emitting `description: null` for automations that have
+    # none would rewrite every automation.yaml (and its content hash) on the
+    # first cycle, pushing one no-op commit per already-synced automation.
+    if automation.description is not None:
+        fields["description"] = automation.description
     return fields
 
 
--- a/openhands/automation/models.py
+++ b/openhands/automation/models.py
@@ -61,6 +61,10 @@ class Automation(Base):
         String(256), nullable=True
     )
 
+    # Optional human-facing description, shown wherever the automation is
+    # listed. Falls back to the prompt in the UI when absent.
+    description: Mapped[str | None] = mapped_column(Text, nullable=True)
+
     # Optional prompt (set when created via preset endpoints)
     prompt: Mapped[str | None] = mapped_column(Text, nullable=True)
 
--- a/openhands/automation/preset_router.py
+++ b/openhands/automation/preset_router.py
@@ -138,6 +138,14 @@ class CreatePromptAutomationRequest(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
     name: str = Field(..., min_length=1, max_length=500)
+    description: str | None = Field(
+        default=None,
+        max_length=2000,
+        description=(
+            "Optional human-facing description, shown wherever the automation "
+            "is listed. Falls back to the prompt in the UI when absent."
+        ),
+    )
     prompt: str = Field(
         ...,
         min_length=1,
@@ -531,6 +539,7 @@ async def create_automation_from_prompt(
             user_id=user.user_id,
             org_id=user.org_id,
             name=body.name,
+            description=body.description,
             prompt=body.prompt,
             preset_metadata=preset_metadata,
             model=model,
@@ -618,6 +627,14 @@ class CreatePluginAutomationRequest(BaseModel):
     model_config = ConfigDict(extra="forbid")
 
     name: str = Field(..., min_length=1, max_length=500)
+    description: str | None = Field(
+        default=None,
+        max_length=2000,
+        description=(
+            "Optional human-facing description, shown wherever the automation "
+            "is listed. Falls back to the prompt in the UI when absent."
+        ),
+    )
     plugins: list[PluginSource] | None = Field(
         default=None,
         description="Plugin(s) to load. Mutually exclusive with 'variants'.",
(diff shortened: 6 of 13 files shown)


santhiprakash:

PR #390, opened 2026-08-26
fix: honour the stored setup_script_path when executing automations

### LINKED ISSUE #343: setup_script_path is validated, stored and tested — but never read; only a root setup.sh ever runs
`setup_script_path` is accepted, validated, persisted and covered by tests, but no executor ever reads it. Both execution paths run a hardcoded `setup.sh` at the tarball root instead, so a custom tarball that names its setup script anything else silently skips its setup step and then fails in the entrypoint.

## What happens

The field is honoured everywhere except where it matters:

| Stage | Behaviour |
| --- | --- |
| Request | `CreateAutomationRequest.setup_script_path` validated by `_validate_command_string` — relative, no `..`, no shell metacharacters (`schemas.py:270`, `:299-302`) |
| Persistence | Stored on the model (`router.py:93`, `models.py:81`, `migrations/versions/001_initial_schema.py:53`) |
| Response | Echoed back in `AutomationResponse` (`schemas.py:626`) |
| Execution | **Never read.** Both paths run `... && cd {work_dir} && ([ ! -f setup.sh ] || bash setup.sh) && {entrypoint}` (`execution.py:416`, `execution.py:581`) |

`grep -rn setup_script_path openhands/` returns only the request models, the router, and the ORM model — no executor, no backend.

## Why it is worth fixing rather than leaving

- **The failure is silent.** The setup step is skipped, the run proceeds, and the error surfaces later as a `ModuleNotFoundError` (or similar) from the entrypoint — with nothing in the logs pointing at the skipped setup.
- **The test suite blesses a value that cannot work.** `tests/test_router.py:436-441` — *"Valid setup_script_path is accepted"* — asserts a 2xx for `"scripts/setup.sh"`, a nested path the runner will never execute.
- **The documentation promises it works.** `skills/openhands-automation/references/custom-automation.md` in OpenHands/extensions documents `setup_script_path` as the way to name the setup script, and its worked example passes `"setup.sh"` alongside an entrypoint of `.venv/bin/python main.py`.
- **The presets are unaffected, which is why this has gone unnoticed.** They hardcode `setup_script_path="setup.sh"` (`preset_router.py:596`, `:1025`), matching the literal the runner looks for. Only user-authored tarballs hit the gap.

## Possible resolutions

**Honour the field.** Use the stored value when set, keeping the current default:

```python
setup_cmd = automation.setup_script_path or "setup.sh"
# ... && ([ ! -f "{setup_cmd}" ] || bash "{setup_cmd}") && {entrypoint}
```

The value is already validated against traversal, absolute paths and shell metacharacters, so interpolation is safe — quoting it keeps it that way.

**Or drop it.** Reject anything but `"setup.sh"` at the request layer, document the root-`setup.sh` convention, and deprecate the column.

Either way the fix should update `tests/test_router.py` (the nested-path case currently asserts the wrong thing) and the extensions documentation.

Found while designing catalog-shipped tarball automations in OpenHands/extensions, where the entry has to declare how its bundle is set up.

santhiprakash:
## Problem

`setup_script_path` is accepted, validated, persisted and echoed back by the API, but no executor ever reads it. Both execution paths build `... && cd {work_dir} && ([ ! -f setup.sh ] || bash setup.sh) && {entrypoint}` with a hardcoded root `setup.sh` (`execution.py:416`, `execution.py:581`), so a tarball that names its setup script anything else silently skips its setup step and then fails in the entrypoint (typically `ModuleNotFoundError`) with nothing in the logs pointing at the skipped setup. The presets are unaffected only because they hardcode `setup_script_path="setup.sh"`.

Fixes #343 (option "Honour the field" from the issue).

## Triage / Root cause

`grep -rn setup_script_path openhands/` on `main` returns only the request models (`schemas.py`), the router pass-through (`router.py:137`), the ORM model (`models.py:84`), the git-sync serializer and the preset constructions — the two `cmd` builders in `execution.py` never see the stored value, because `execute_in_context` does not receive it (the dispatcher has the `Automation` row in scope at its call site but only passes `entrypoint`).

## Fix

- `execute_in_context` takes a new `setup_script_path: str | None = None` parameter and interpolates it into the setup fragment with the existing `_shell_quote` helper, keeping the `setup.sh` default when unset: `([ ! -f '{path}' ] || bash '{path}')`. The value is already validated at the request layer (relative, no `..` segments, no shell metacharacters), and single-quoting keeps a `'` in a stored path literal.
- `run_automation` (blocking path, used by the e2e scripts) gains the same parameter and interpolation.
- The dispatcher passes `setup_script_path=automation.setup_script_path` at its single production call site.
- Docstrings that promised `setup.sh` now describe the configured setup script.

## Verification

- `uv run pytest tests/ -q --ignore=tests/integration` → **1460 passed** (includes 4 new tests in `tests/test_execution.py::TestSetupScriptPath`: custom path reaches the dispatch command, default `setup.sh` when unset, a single quote in the path stays escaped, and the blocking path honours the stored path).
- Rebased onto `main` @ `3efb91d` (`020_add_automation_disabled_reason` landed after opening): full suite → **1473 passed, 0 failed**; the dispatcher hunk composes cleanly with the new unhealthy-automation logic.
- Rebased onto `main` @ `f1b3244` (release 1.9.0, includes SDK 1.44.0 bump #393; no hunk overlap with dispatcher/execution): targeted suite `tests/test_execution.py` + `tests/test_dispatcher.py` → **85 passed, 0 failed** on SDK 1.44.0; pre-commit (ruff format/lint, pycodestyle, pyright) all pass.
- `uv run pre-commit run --files openhands/automation/execution.py openhands/automation/dispatcher.py tests/test_execution.py` → ruff format/lint, pycodestyle, pyright all passed.

## Notes

- `tests/test_router.py`'s "valid setup_script_path is accepted" case (asserting 201 for `"scripts/setup.sh"`) needed no change: the issue called it out as blessing a value the runner would never execute, and after this fix that value is genuinely executed, so the assertion is now correct end-to-end.
- Open PR #179 (cross-platform preset bootstrap) also threads `setup_script_path` into `execute_in_context` as part of a larger rework. That PR is aimed at a different problem (POSIX-free presets), has changes requested (duplicated bootstrap + Windows `os.execv`), and has been open since June, so I did not treat it as covering this issue — but if it revives, this change will need a small rebase on the same lines.
- Presets keep `setup_script_path="setup.sh"` and are unaffected; git-synced `automation.yaml` values round-trip through the existing serializer and now take effect on the next run.

HUMAN: This PR was prepared with AI assistance (code search, patch, and test execution guided by a human reviewer); the fix follows the approach sketched in issue #343.

### DIFF
--- a/openhands/automation/dispatcher.py
+++ b/openhands/automation/dispatcher.py
@@ -421,6 +421,7 @@ async def _execute_run(
             env_vars=env_vars,
             timeout=effective_timeout,
             run_id=run_id,
+            setup_script_path=automation.setup_script_path,
             sandbox_id=ctx.sandbox_id,
         )
     except PermanentDispatchError as exc:
--- a/openhands/automation/execution.py
+++ b/openhands/automation/execution.py
@@ -341,6 +341,7 @@ async def execute_in_context(
     timeout: int | None = None,
     run_id: str | None = None,
     sandbox_id: str | None = None,
+    setup_script_path: str | None = None,
 ) -> DispatchResult:
     """Execute automation code in an existing execution context.
 
@@ -348,7 +349,7 @@ async def execute_in_context(
     The context (agent_url, session_key) is obtained from the backend.
 
     1. Get tarball into environment (upload bytes OR download from URL).
-    2. Extract it, run ``setup.sh`` (if present), then start *entrypoint*.
+    2. Extract it, run the setup script (if present), then start *entrypoint*.
     3. Return immediately without waiting for the entrypoint to complete.
 
     Args:
@@ -364,6 +365,10 @@ async def execute_in_context(
             path (/tmp/automation-<run_id>.tar.gz) that prevents collisions
             when concurrent runs share the same filesystem (sandboxless mode)
         sandbox_id: Sandbox ID for logging (Cloud mode only)
+        setup_script_path: Path to the setup script inside the extracted
+            tarball (default: ``setup.sh``). The value is validated by the
+            request layer (relative, no traversal, no shell metacharacters)
+            before it is stored on the automation.
 
     Returns:
         DispatchResult with success status
@@ -408,12 +413,13 @@ async def execute_in_context(
             )
             env_prefix = _env_command_prefix(env_path)
 
+        setup_cmd = _shell_quote(setup_script_path or "setup.sh")
         cmd = (
             f"{env_prefix}mkdir -p {work_dir}"
             f" && tar xzf {tarball_path} -C {work_dir}"
             f" && rm -f {tarball_path}"
             f" && cd {work_dir}"
-            f" && ([ ! -f setup.sh ] || bash setup.sh)"
+            f" && ([ ! -f {setup_cmd} ] || bash {setup_cmd})"
             f" && {entrypoint}"
         )
 
@@ -484,6 +490,7 @@ async def run_automation(
     run_id: str | None = None,
     keep_sandbox: bool = False,
     work_dir: str = DEFAULT_WORK_DIR,
+    setup_script_path: str | None = None,
 ) -> AutomationResult:
     """Execute an automation end-to-end in a fresh sandbox (blocking).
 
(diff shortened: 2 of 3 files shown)


santhiprakash:

Issue #347, opened 2026-08-19
Run summaries are only readable one automation at a time, and the SQLite engine's undeclared pool turns the resulting fan-out into 30s stalls + HTTP 500


VascoSch92:
Split out from OpenHands/OpenHands#16559, where the Agent Canvas Automation view
takes ~30s to load. That issue keeps the frontend half (bounded fan-out, sidebar
polling). This one is the automation-service half.

## What the client is forced to do

`GET /v1` (`router.py:148`) returns bare automation rows, and there is no
endpoint anywhere in the service that returns run data for more than one
automation — `latest_run`, `last_run`, `summary` and `summaries` have no
occurrences in `router.py` or `schemas.py`. Any surface that shows per-automation
health therefore has to issue one `GET /v1/{id}/runs` per listed automation.
The list permits `limit=100`, so the client-side fan-out is bounded only by the
page size.

Each of those requests costs three DB reads (`router.py:381-397`): the ownership
lookup in `_get_user_automation`, a `COUNT(*)` over the automation's runs, and
the paginated page itself. The reporter's 21 automations therefore cost 1 + 21
requests and 2 + 63 queries to draw one screen.

In the HAR on OpenHands#16559: 21 concurrent `GET /v1/{id}/runs?limit=20`, all
starting within ~5ms of each other, all taking 30.3-30.8s, 15 returning 200 and
6 returning 500. Immediate retries of the failures succeeded in 295-450ms.

### Fix

Serve the aggregate the clients are currently simulating — either

```
GET /v1?limit=50&include=latest_run
```

or a dedicated batch read:

```
GET /v1/runs/summaries?automation_ids=...&sample_limit=20
```

One authorization query over all requested IDs, one windowed query for the
newest N runs per automation
(`ROW_NUMBER() OVER (PARTITION BY automation_id ORDER BY created_at DESC)`,
supported on SQLite 3.25+ as well as Postgres), one grouped `COUNT(*)`. Bound
the ID count and the sample size. Returning run rows rather than computed health
labels keeps the current frontend-owned dashboard semantics intact.

This turns 21 requests into 1 and is the only change here that also holds up
with several concurrent users — a client-side concurrency cap is per browser tab
and does not compose.

## The engine does not have the pool it says it has

Independent of the N+1, the SQLite path silently gets a pool it documents itself
as not having. `_create_sqlite_engine` (`db.py:151-170`):

```python
    # No pooling for SQLite - it handles this internally
    engine = create_async_engine(
[...]

PR #391, opened 2026-08-26
fix(db): use NullPool for SQLite to eliminate silent connection pooling (#347)

santhiprakash:
## Problem

`_create_sqlite_engine()` claimed "No pooling for SQLite - it handles this internally" but passed `pool_pre_ping=True` without specifying a `poolclass`. SQLAlchemy therefore applied its default `AsyncAdaptedQueuePool` (5 connections, 10 overflow, 30 s timeout). Under concurrent fan-out (e.g. the Automation view fetching runs for 21 automations), this pool silently exhausted and produced HTTP 500s on 6 of 21 concurrent requests (#347).

The `db_pool_size`, `db_max_overflow`, and `db_pool_timeout` config values were also silently ignored by the SQLite path.

## Root cause

`create_async_engine()` without an explicit `poolclass` uses the default `AsyncAdaptedQueuePool` regardless of the comment claiming otherwise. The misleading `pool_pre_ping=True` reinforced the wrong mental model.

## Fix

- Pass `poolclass=NullPool` to `create_async_engine()` in `_create_sqlite_engine()`, which disables SQLAlchemy-level pooling entirely — each request opens and closes its own connection, matching the documented intent.
- Remove `pool_pre_ping=True` (no-op with `NullPool`).
- Update the docstring to reference #347 and explain the `NullPool` choice.
- Add a regression test (`test_uses_null_pool`) asserting the engine's pool class is `NullPool`.

The N+1 fan-out problem (separate batch-read endpoint) is a separate fix tracked in the same issue; this PR addresses only the pool configuration bug.

## Verification

```
uv run pytest tests/ -q --ignore=tests/integration
→ 1457 passed, 0 failed

uv run pre-commit run --files openhands/automation/db.py tests/test_db.py
→ ruff format, ruff lint, pycodestyle, pyright: all Passed
```

## Notes / Risks

- `NullPool` is the standard SQLAlchemy recommendation for SQLite in multi-threaded/async contexts (connections are lightweight and SQLite handles its own internal locking).
- The Postgres and GCP Cloud SQL paths are unchanged — they continue to use `AsyncAdaptedQueuePool` with the configured pool settings.
- Addresses only the pool half of #347; the aggregate/batch-read endpoint to eliminate the N+1 client-side fan-out is a separate PR.

HUMAN: This PR was authored with AI assistance.


santhiprakash:
Cross-linking #382 since it touches the same `_create_sqlite_engine()` path with the opposite direction on the same bug: it honors `db_pool_size` / `db_max_overflow` / `db_pool_timeout` for file-backed SQLite, while this PR disables SQLAlchemy-level pooling entirely.

Trade-off summary for whoever reviews first (defaults verified against the pinned SQLAlchemy 2.0.49):

- The failure mode in #347 was pool **exhaustion** under fan-out (6/21 concurrent requests → 500 after ~30 s). Honoring the configured limits on SQLite keeps that failure mode, just with configurable numbers; `NullPool` removes it and matches the function's documented intent ("no pooling for SQLite").
- In #382's favor: it preserves SQLAlchemy's `StaticPool` default for `:memory:` URLs, where a shared connection is required (with `NullPool`, each new connection is a fresh empty database). This PR's unconditional `NullPool` changes that edge too — not reachable in production today (`db_url` defaults to empty/Postgres, and local mode uses a file path), but it's a fair difference. Happy to add a `:memory:` guard here if that direction is preferred.
- Whichever merges first, the other needs a rework of the same hunk — glad to rebase this onto whichever direction maintainers pick.

santhiprakash:

PR #392, opened 2026-08-27
fix(presets): pin SDK install to the run venv so UV_PYTHON cannot redirect it

### LINKED ISSUE #338: setup.sh installs the SDK outside the run venv when UV_PYTHON is set — silent success, then ModuleNotFoundError in main.py
## Summary

`presets/*/setup.sh` creates a per-run venv and then installs the SDK **without telling `uv` to target that venv**. Because `uv pip install` honours `UV_PYTHON` ahead of the `.venv` in the CWD, any `UV_PYTHON` present in the agent-server's environment silently redirects the install into a *different* environment. `setup.sh` then exits 0, the `&&` chain proceeds, and the entrypoint (`.venv/bin/python main.py`) runs against an empty venv:

```
Traceback (most recent call last):
  File ".../workspaces/automation-runs/<run-id>/main.py", line 136, in <module>
    from openhands.sdk import Conversation, RemoteConversation
ModuleNotFoundError: No module named 'openhands'
```

The failure message points at `main.py`, but nothing is wrong with `main.py` — the install went somewhere else and setup reported success.

Observed with `openhands-automation` 1.7.1 (also present on `main`), agent-server 1.42.1, `uv` 0.9.29, macOS arm64.

## Root cause

The two commands in `setup.sh` resolve to **different interpreters**:

```sh
# presets/prompt/setup.sh:40  — explicit CLI flag, overrides UV_PYTHON
uv venv .venv --python '>=3.12' --quiet

# presets/prompt/setup.sh:43  — no --python, so UV_PYTHON wins over ./.venv
uv pip install --quiet "openhands-sdk==${SDK_VERSION}" ...
```

With `UV_PYTHON=3.13` exported in the environment the agent-server inherits (it reaches the sandbox shell — verified via `/api/bash/execute_bash_command`), `uv pip install -v` reports:

```
DEBUG Found cpython-3.14.2+freethreaded at ./.venv/bin/python3 (virtual environment)
DEBUG Found cpython-3.13.11-macos-aarch64-none at .../archive-v0/<id>/bin/python3.13
DEBUG Using Python 3.13.11 environment at: /Users/<user>/.cache/uv/archive-v0/<id>
Audited 1 package
```

`.../archive-v0/<id>` is the **running agent-server's own uvx environment**. The SDK is already installed there, so the install is a no-op, exits 0, and `setup.sh` prints `[setup] Done`.

## Second-order damage: it writes into the agent-server's environment

This is not merely a misdirected no-op. When the requested `SDK_VERSION` differs from what the agent-server has, the run **overwrites the live server's SDK**. On our box the agent-server is launched as:

```
uvx --from openhands-agent-server==1.42.1 --with openhands-sdk==1.42.1 --with openhands-tools==1.42.1 ...
```

but `/server_info` was reporting `sdk_version: 1.40.1`, and `openhands_sdk-1.40.1.dist-info` in that environment carried the mtime of an automation run. Every run had been quietly downgrading the server it was talking to. A clean restart brought it back to 1.42.1.

## Secondary bug: `--python '>=3.12'` can select a free-threaded interpreter

On a machine with free-threaded CPython installed, `uv venv --python '>=3.12'` picks it (it is simply the highest match):

```
$ uv venv .venv --python '>=3.12'
Using CPython 3.14.2+freethreaded
```

The SDK cannot be installed there at all:

```
× Failed to download and build `tree-sitter-bash==0.25.1`
╰─▶ The built wheel `tree_sitter_bash-0.25.1-cp310-abi3-macosx_11_0_arm64.whl`
    is not compatible with the current Python 3.14 on macos aarch64
```

So even with the `UV_PYTHON` redirect fixed, setup would fail on such a host — just loudly instead of silently. The `>=3.12` request should exclude free-threaded builds (and probably cap the upper bound to versions the SDK ships wheels for).

## Reproduction

1. Start a stack with `UV_PYTHON` set to any version that exists somewhere on the host other than the run venv (e.g. `UV_PYTHON=3.13 npm run dev` in agent-canvas).
2. Trigger any automation.
3. The run fails with `ModuleNotFoundError: No module named 'openhands'` while stdout shows `[setup] Done`.
4. `ls <run-dir>/.venv/lib/python*/site-packages` contains only `_virtualenv.pth` — nothing was installed.

## Suggested fix

Make the install target the venv explicitly, so no ambient environment variable can redirect it:

```sh
uv venv .venv --python '>=3.12' --quiet
uv pip install --python .venv/bin/python --quiet \
  "openhands-sdk==${SDK_VERSION}" ...
```

(or export `VIRTUAL_ENV="$PWD/.venv"` and `unset UV_PYTHON` before the install).

Two hardening suggestions alongside it:

- Verify the install before handing off, so a broken setup fails at the setup step instead of surfacing as a confusing traceback in `main.py`:
  `.venv/bin/python -c 'import openhands.sdk' || { echo "[setup] ERROR: SDK not importable in .venv" >&2; exit 1; }`
- Constrain the interpreter request so a free-threaded build is never chosen.

Both preset copies are affected — `presets/prompt/setup.sh` and `presets/plugin/setup.sh` are byte-identical.

santhiprakash:
## Problem

Preset `setup.sh` creates a per-run venv and then installs the OpenHands SDK with a bare `uv pip install`. `uv` honours an ambient `UV_PYTHON` **ahead of the `.venv` in the CWD**, so when the agent-server's environment carries `UV_PYTHON` (e.g. it was launched via `uvx`/`uv run`), the install is silently redirected into a *different* environment: `setup.sh` prints `[setup] Done` and exits 0, the run venv stays empty, and the entrypoint fails with a confusing traceback that points at `main.py`:

```
ModuleNotFoundError: No module named 'openhands'
```

Reproduced locally with uv 0.11.7: with a 3.12 `.venv` in the CWD and `UV_PYTHON` pointing at a 3.13 environment, `uv pip install six` reports `Using Python 3.13.13 environment at: fake-server-env` — `six` lands in the 3.13 env, `.venv/bin/python -c "import six"` fails. Exactly the mechanism in #338 (which also documents the second-order damage: when the SDK version differs, the run *overwrites the live agent-server's* environment).

Closes #338.

## Triage / Root cause

`openhands/automation/presets/prompt/setup.sh:43` (and the byte-identical `presets/plugin/setup.sh` + `scripts/test_tarball/setup.sh`):

```sh
uv venv .venv --python '>=3.12' --quiet   # explicit flag: UV_PYTHON ignored here
uv pip install --quiet ...                # no flag: ambient UV_PYTHON wins over ./.venv
```

## Fix

- `unset UV_PYTHON` before the `uv pip install` in all three copies of the script, so no ambient value can redirect the install away from the run venv.
- After the install, verify the SDK is importable from the run venv and fail loudly at the setup step (covering both POSIX `.venv/bin/python` and Windows `.venv/Scripts/python.exe` layouts — the entrypoint already supports both via `_get_preset_entrypoint`), so a misdirected install surfaces as `[setup] ERROR: ...` instead of a `ModuleNotFoundError` inside `main.py`.

I used `unset UV_PYTHON` rather than the `--python .venv/bin/python` variant sketched in the issue because hardcoding the POSIX venv path would break Windows sandboxes; with `UV_PYTHON` cleared, `uv pip install` resolves to the `.venv` in the CWD (which `execution.py` guarantees is the run's work dir) portably.

The secondary `--python '>=3.12'` free-threaded-interpreter issue from #338 is **not** addressed here (needs an interpreter-request policy decision) — happy to follow up separately.

## Verification

- Reproduced the redirect (old behavior) and the fix (new fragment installs into `.venv`; the import check fails loudly on an empty venv) — sequence quoted in the issue monitor notes.
- `bash -n` clean on all three copies.
- `uv run pytest tests/ -q --ignore=tests/integration` → **1463 passed**, including 8 new regression tests asserting both preset copies (and the test-tarball copy) clear `UV_PYTHON` and verify the SDK import, and that the verification handles both venv layouts.
- `uv run pre-commit run --files <changed>` → ruff format/lint, pycodestyle, pyright all pass.

## Notes / Risks

- Complementary to (not overlapping) open PR #349, which unsets an inherited `VIRTUAL_ENV` for the same failure class (`uv` also honours `VIRTUAL_ENV`); the two changes compose cleanly. If #349 merges first this applies on top with at most a trivial rebase.
- Pre-existing content tests (`test_*_setup_sh_fetches_sdk_version_from_api`, `test_test_tarball_setup_fetches_service_sdk_version`) pass unchanged.
- Maintainer-authored issue (#338, VascoSch92).

HUMAN: Prepared by Santhi Prakash with AI assistance (code, verification, PR description); reviewed and submitted by Santhi Prakash.

### DIFF
--- a/openhands/automation/presets/plugin/setup.sh
+++ b/openhands/automation/presets/plugin/setup.sh
@@ -40,9 +40,32 @@ echo "[setup] Creating isolated virtual environment"
 uv venv .venv --python '>=3.12' --quiet
 
 echo "[setup] Installing OpenHands SDK from PyPI (version: $SDK_VERSION)"
+# Clear UV_PYTHON so an ambient value (e.g. exported by the agent-server's
+# own uvx/uv launcher) cannot redirect the install away from the .venv we
+# just created in the CWD: `uv pip install` honours UV_PYTHON ahead of the
+# local .venv, silently installing into a different environment while the
+# run venv stays empty (#338).
+unset UV_PYTHON
 uv pip install --quiet \
   "openhands-sdk==${SDK_VERSION}" \
   "openhands-tools==${SDK_VERSION}" \
   "openhands-workspace==${SDK_VERSION}"
+
+# Fail at the setup step when the SDK is not importable from the run venv:
+# a misdirected install would otherwise surface later as a confusing
+# ModuleNotFoundError inside main.py (#338).
+echo "[setup] Verifying SDK is importable from the run venv"
+if [ -x .venv/bin/python ]; then
+    VENV_PYTHON=.venv/bin/python
+elif [ -x .venv/Scripts/python.exe ]; then
+    VENV_PYTHON=.venv/Scripts/python.exe
+else
+    echo "[setup] ERROR: run venv python not found in .venv" >&2
+    exit 1
+fi
+"$VENV_PYTHON" -c 'import openhands.sdk, openhands.tools.preset, openhands.workspace' || {
+    echo "[setup] ERROR: OpenHands SDK not importable in .venv (install misdirected?)" >&2
+    exit 1
+}
 
 echo "[setup] Done"
--- a/openhands/automation/presets/prompt/setup.sh
+++ b/openhands/automation/presets/prompt/setup.sh
@@ -40,9 +40,32 @@ echo "[setup] Creating isolated virtual environment"
 uv venv .venv --python '>=3.12' --quiet
 
 echo "[setup] Installing OpenHands SDK from PyPI (version: $SDK_VERSION)"
+# Clear UV_PYTHON so an ambient value (e.g. exported by the agent-server's
+# own uvx/uv launcher) cannot redirect the install away from the .venv we
+# just created in the CWD: `uv pip install` honours UV_PYTHON ahead of the
+# local .venv, silently installing into a different environment while the
+# run venv stays empty (#338).
+unset UV_PYTHON
 uv pip install --quiet \
   "openhands-sdk==${SDK_VERSION}" \
   "openhands-tools==${SDK_VERSION}" \
   "openhands-workspace==${SDK_VERSION}"
+
+# Fail at the setup step when the SDK is not importable from the run venv:
+# a misdirected install would otherwise surface later as a confusing
+# ModuleNotFoundError inside main.py (#338).
+echo "[setup] Verifying SDK is importable from the run venv"
+if [ -x .venv/bin/python ]; then
+    VENV_PYTHON=.venv/bin/python
+elif [ -x .venv/Scripts/python.exe ]; then
+    VENV_PYTHON=.venv/Scripts/python.exe
+else
+    echo "[setup] ERROR: run venv python not found in .venv" >&2
+    exit 1
+fi
+"$VENV_PYTHON" -c 'import openhands.sdk, openhands.tools.preset, openhands.workspace' || {
+    echo "[setup] ERROR: OpenHands SDK not importable in .venv (install misdirected?)" >&2
+    exit 1
+}
 
 echo "[setup] Done"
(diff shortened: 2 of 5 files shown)


santhiprakash:
Closing for now to respect review bandwidth — we already have a few PRs queued in this repo. The branch stays up and the fix is complete; happy to reopen once the others land. Thanks!

PR #395, opened 2026-08-27
feat: add pinned_at field to Automation for cross-device pin sync

### LINKED ISSUE #394: feat: add pinned/pinned_order fields to Automation so pin state syncs across devices
## Problem

Pinned automations on the Agent Canvas home dashboard are stored only in browser localStorage, so pin state and order do not follow a user across devices, browsers, or profiles. The Canvas-side migration is tracked in OpenHands/OpenHands#16768.

### Desired Behavior

Persist pin state with each automation record:

- `pinned_at` is nullable; null means unpinned and a UTC timestamp means pinned.
- Pinning updates `pinned_at` to the current UTC time, allowing pinned automations to be ordered by recency.
- Unpinning clears `pinned_at`.
- Existing automation records remain unpinned after migration.

The list and detail APIs expose `pinned_at`, and the existing update endpoint accepts a pin/unpin operation. Canvas can then derive pin state and order from backend data and migrate its existing local pins.

### Acceptance Criteria

- [ ] The Automation model has a nullable `pinned_at` UTC timestamp that defaults to null for existing and new unpinned records.
- [ ] Automation list and detail responses include `pinned_at`.
- [ ] `PATCH /v1/automations/{id}` with `{"pinned": true}` sets `pinned_at` to the current UTC time, including when re-pinning an already pinned automation to change its order.
- [ ] `PATCH /v1/automations/{id}` with `{"pinned": false}` clears `pinned_at`.
- [ ] An Alembic migration adds the field without changing existing automation behavior.
- [ ] API tests cover pinning, re-pinning, unpinning, persistence, and response serialization.

### Related

- OpenHands/OpenHands#16768
- OpenHands/automation#395

_This issue was updated by an AI agent (OpenHands) on behalf of Gregory Neubig._

BSmick6:
## Summary

Closes #394

Pinned automations are currently stored in browser \`localStorage\`, so pin state is lost when switching devices or browsers. This PR persists pin state in the database via a new nullable \`pinned_at\` column.

- **\`Automation\` model** (\`openhands/automation/models.py\`): adds \`pinned_at: DateTime(timezone=True), nullable=True\`. \`NULL\` = unpinned; a UTC timestamp = pinned (ordering by recency gives pin order).
- **\`AutomationResponse\` schema** (\`openhands/automation/schemas.py\`): exposes \`pinned_at: UtcDatetime | None\` so all list and detail endpoints return the field.
- **\`UpdateAutomationRequest\` schema** (\`openhands/automation/schemas.py\`): adds \`pinned: bool | None\`. The PATCH handler converts \`true\` → current UTC timestamp and \`false\` → \`NULL\` before writing to the model, so callers never deal with timestamps directly.
- **Router** (\`openhands/automation/router.py\`): translates \`pinned\` → \`pinned_at\` in \`update_automation\` before the generic field-copy loop.
- **Migration** (\`migrations/versions/022_add_automation_pinned_at.py\`): Alembic migration that \`ADD COLUMN pinned_at TIMESTAMPTZ NULL\` with \`NULL\` defaults for all existing rows.

## HUMAN: Testing

Ran the service locally against a fresh PostgreSQL instance with migrations applied. Created an automation via \`POST /api/automation/v1\` and confirmed \`pinned_at\` was \`null\`. Sent \`PATCH {"pinned": true}\` and verified the response contained a UTC timestamp; a subsequent GET confirmed it persisted. Sent \`PATCH {"pinned": false}\` and verified \`pinned_at\` returned to \`null\` on both the response and a follow-up GET.

---
🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/openhands/automation/router.py
+++ b/openhands/automation/router.py
@@ -226,6 +226,10 @@ async def update_automation(
     auto = await _get_user_automation(session, automation_id, user.user_id, user.org_id)
 
     update_data = body.model_dump(exclude_unset=True)
+    # Convert pinned boolean to pinned_at timestamp before writing to the model
+    if "pinned" in update_data:
+        pinned = update_data.pop("pinned")
+        update_data["pinned_at"] = utcnow() if pinned else None
     # Handle trigger field mapping (only if trigger has a real value)
     if body.trigger is not None:
         update_data["trigger"] = body.trigger.model_dump()
--- a/migrations/versions/022_add_automation_pinned_at.py
+++ b/migrations/versions/022_add_automation_pinned_at.py
@@ -1,0 +1,41 @@
+"""Add pinned_at column to automations.
+
+Revision ID: 022
+Revises: 021
+Create Date: 2026-08-27
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+
+revision: str = "022"
+down_revision: str = "021"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def _is_sqlite() -> bool:
+    return op.get_bind().dialect.name == "sqlite"
+
+
+def upgrade() -> None:
+    op.add_column(
+        "automations",
+        sa.Column("pinned_at", sa.DateTime(timezone=True), nullable=True),
+    )
+
+    if _is_sqlite():
+        return
+
+    op.execute(
+        "COMMENT ON COLUMN automations.pinned_at IS "
+        "'Timestamp when the automation was pinned. NULL means unpinned. "
+        "Ordered by recency to determine pin order.'"
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("automations", "pinned_at")
--- a/openhands/automation/models.py
+++ b/openhands/automation/models.py
@@ -104,6 +104,11 @@ class Automation(Base):
         DateTime(timezone=True), nullable=True
     )
 
+    # Pinned timestamp (NULL = not pinned; set to UTC now when pinned)
+    pinned_at: Mapped[datetime | None] = mapped_column(
+        DateTime(timezone=True), nullable=True
+    )
+
     # Soft delete timestamp (NULL = not deleted)
     deleted_at: Mapped[datetime | None] = mapped_column(
         DateTime(timezone=True), nullable=True, index=True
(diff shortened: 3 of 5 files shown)


openhands/automation/router.py
```diff
@@ -226,6 +226,10 @@ async def update_automation(
     auto = await _get_user_automation(session, automation_id, user.user_id, user.org_id)
 
     update_data = body.model_dump(exclude_unset=True)
+    # Convert pinned boolean to pinned_at timestamp before writing to the model
+    if "pinned" in update_data:
+        pinned = update_data.pop("pinned")
+        update_data["pinned_at"] = utcnow() if pinned else None
```
BSmick6, line 232: This always refreshes `pinned_at` even if the automation is already pinned, so the semantics are "ordered by most recent pin action" rather than "ordered by original pin time." Two behaviors fall out of this:

- **"Move to top" works in one request** — sending `pinned: true` to an already-pinned automation bumps it to the front of the pin order.
- **A bulk-save that includes `pinned: true` for all pinned items silently reorders them** to the save timestamp.

If "move to top" is intentional, this is correct and worth a brief comment here. If stable ordering (first-pinned = front) is the intent, guard with:

```python
if pinned:
    update_data["pinned_at"] = auto.pinned_at or utcnow()
else:
    update_data["pinned_at"] = None
```

Which behavior is intended?

---
_Generated by [Claude Code](https://claude.ai/code)_

BSmick6:
Marking as draft — the pinned_at timestamp approach can’t support arbitrary drag reorder (as designed in https://github.com/OpenHands/OpenHands/issues/16768). Sorting by pin time works for “most recently pinned first” but can’t express a user-defined position between two existing pins. Need to settle the data model on the issue before landing this.

BSmick6:
Closing this in favour of a frontend-only approach.

After reviewing the overall architecture, we decided to persist pin state in `misc_settings.ui_preferences.home_pinned_automations` on the agent-server settings API rather than adding a `pinned_at` column to the automation model. The agent-server already has a generic `misc_settings` blob that the canvas frontend owns — using it means:

- No schema migration or new API surface on the automation service.
- Pin state naturally lives per-user/per-frontend (which is correct — pins are a UI preference, not part of the automation itself).
- One-time migration from localStorage to the server happens in the frontend hook on first load.

The canvas-side implementation landed in [OpenHands/OpenHands#17104](https://github.com/OpenHands/OpenHands/pull/17104), which closes [OpenHands/OpenHands#16768](https://github.com/OpenHands/OpenHands/issues/16768).

Closing this PR and issue #394.

Issue #396, opened 2026-08-27
Automations are invisible and unmanageable to other members of the org they run for


VascoSch92:
## Summary

Managing an automation is scoped to `(user_id, org_id)`, but *running* one is scoped to `org_id` alone — and the scheduler is scoped to neither. An automation created by one account therefore fires on every event for its whole org, while being invisible and unmanageable to every other member of that org.

## Where

| Path | Scoping |
|---|---|
| `GET /automations` (`router.py`, `list_automations`) | `user_id == caller` **AND** `org_id == caller's org` AND `deleted_at IS NULL` |
| get / patch / delete (`router.py`, `_get_user_automation`) | same filter → **404** for any other member |
| `POST /automations/runs/{run_id}/cancel` (`router.py`, `cancel_run`) | **403** `"Not your automation"` |
| event matching (`utils/webhook.py`, `find_matching_automations`) | `org_id` + `enabled` + `deleted_at` — **no `user_id`** |
| scheduler claim (`scheduler.py`) | `enabled` + `deleted_at` only — **no org, no user** |

## Impact

When a shared or service account owns an org's automations, no other member of that org can see, disable or delete them — including org admins. There is no org-wide listing endpoint (every route in `router.py` is owner-scoped), so the automation simply does not appear anywhere in the UI while it continues to act on the org's repositories and post under the org's bot identity.

The failure mode is particularly confusing: because the automation is absent from the automations page, it reads as "already deleted" while it is in fact still running. The only remedies today are authenticating as the exact owning account with that org selected, or writing to the database directly.

## Diagnosing this is harder than it should be

`last_triggered_at` is written only by the polling path (`utils/run.py`, `create_pending_run`). The event path (`utils/webhook.py`, `create_automation_run`) never writes it, so an event-triggered automation reports `last_triggered_at = NULL` forever regardless of how often it fires.

Any query that ranks by that column to find "currently active" automations will silently miss every event-triggered one. `automation_runs.created_at` is the only reliable signal. It would help either to write `last_triggered_at` on the event path too, or to document that it is polling-only.

## Reproduction

1. Org `O` with members `A` and `B`.
2. `A` creates an event-triggered automation in `O`.
3. As `B` — any role, including org admin — with `O` selected:
   - `GET /automations` → the automation is not listed
   - `DELETE /automations/{id}` → `404 Automation not found`
   - `POST /automations/runs/{run_id}/cancel` → `403 Not your automation`
[...]

PR #399, opened 2026-08-28
fix: scope automation management to the org instead of the owner

VascoSch92:
Fixes the scoping half of #396.

## Problem

Management endpoints filtered on `user_id` **AND** `org_id`, while event matching filtered on `org_id` alone and the scheduler on neither. An automation created by one account therefore fired for its whole org — acting on the org's repos under the org's bot identity — while being invisible and unmanageable to every other member of that org: it was absent from `GET /automations`, `DELETE` returned 404, and cancel returned 403. With no org-wide listing endpoint, it read as "already deleted" while still running.

This is easy to hit with a shared or service account that owns an org's automations, and the only remedies were authenticating as that exact account or writing to the database.

## Change

`list`, `get`, `patch`, `delete`, `tarball`, `dispatch`, `runs` and `cancel_run` now resolve by `org_id`. `_get_user_automation` becomes `_get_org_automation`; the `user_id` predicate is dropped from the list query and from the `cancel_run` ownership check.

Authorization is unchanged: every one of these routes already depends on `require_permission("manage_automations")`, which is org-scoped, so dropping the owner predicate does not leave them ungated. `user_id` is still recorded on create — it just no longer governs access.

Template dedupe (`find_existing_template_automation`) is org-scoped to match, so enabling a template twice in one org returns the existing automation rather than creating a per-member duplicate.

`complete_run` and `report_run_phase` are callbacks from inside the sandbox rather than user actions, so they keep the owner check.

No schema change and no migration — `org_id` is already present and indexed.

## Tests

- `test_list_automations_includes_other_org_members` — a teammate's automation is listed
- `test_delete_other_org_members_automation` — a member can delete a teammate's automation
- `test_cancel_same_org_other_users_run` — a member can cancel a teammate's run
- `test_cancel_other_users_run_returns_403` renamed to `test_cancel_other_orgs_run_returns_403`; it was always cross-org, and cross-org rejection is unchanged

Full unit suite passes (1483). Pre-commit clean: ruff format, ruff lint, pycodestyle, pyright.

## Deliberately not included

The `last_triggered_at` observability problem also described in #396 — the event path never writes it, so event-triggered automations report `NULL` forever and any query ranking by it to find active automations silently misses them. That is a separate concern; keeping this PR to the access-control change.

## Follow-up worth considering

The list response has no owner field, so the UI will now show automations without indicating who created them. A small `agent-canvas` change to surface the owner would help.

### DIFF
--- a/openhands/automation/preset_router.py
+++ b/openhands/automation/preset_router.py
@@ -464,7 +464,7 @@ async def create_automation_from_prompt(
     # the existing automation unchanged instead of creating a duplicate.
     if body.template is not None:
         existing = await find_existing_template_automation(
-            session, user.user_id, user.org_id, body.template.id
+            session, user.org_id, body.template.id
         )
         if existing is not None:
             response.status_code = status.HTTP_200_OK
@@ -871,7 +871,7 @@ async def create_automation_from_plugin(
     # the existing automation unchanged instead of creating a duplicate.
     if body.template is not None:
         existing = await find_existing_template_automation(
-            session, user.user_id, user.org_id, body.template.id
+            session, user.org_id, body.template.id
         )
         if existing is not None:
             response.status_code = status.HTTP_200_OK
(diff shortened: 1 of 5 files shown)


tofarr (approved):
🍰

PR #403, opened 2026-08-28
fix: Forward X-Org-Id during automation auth

### LINKED ISSUE #402: Automation API ignores Canvas-selected X-Org-Id when resolving organization
## Problem

When OpenHands Canvas is served under `/canvas`, the Canvas frontend lets users switch between organization-scoped Cloud backend entries. The automations dashboard refetches after org selection changes, but it can still show automations from the previously resolved organization.

## Root cause

Canvas sends the selected organization as `X-Org-Id` on Cloud requests, including automation requests. The automation service authenticates each request by calling OpenHands `GET /api/v1/users/me`, then uses the returned `org_id` to scope automation queries.

However, the automation service currently forwards only `Authorization` or `Cookie` to `/api/v1/users/me`; it does not forward the incoming `X-Org-Id`. As a result, OpenHands resolves `/users/me` using the user’s persisted/default current org instead of the Canvas-selected org. The automation service then filters with that stale/default org:

```python
Automation.org_id == user.org_id
```

The auth cache is also keyed only by credential, so even if `X-Org-Id` is forwarded, the cache needs to include the requested org scope to avoid reusing one org’s `AuthenticatedUser` for another org.

## Expected behavior

Requests with `X-Org-Id` should authenticate and authorize against that organization, and `/api/automation/v1` should list automations for the selected organization.

## Proposed fix

- Forward incoming `X-Org-Id` when validating credentials with OpenHands `/api/v1/users/me`.
- Include auth method and requested org ID in the automation auth cache key.
- Add tests for `X-Org-Id` forwarding and org-aware cache separation.

malhotra5:
## Summary
- Forward incoming `X-Org-Id` to OpenHands `/api/v1/users/me` during automation API authentication.
- Include auth method and requested org scope in the auth cache key so one credential can authenticate separately per org.
- Add tests for API-key/cookie `X-Org-Id` forwarding, invalid org header handling, and cache separation by org.

Fixes #402

## Testing
- `uv run pytest tests/test_auth.py -q -k 'not TestAuthIntegration'` — passed (50 passed, 4 deselected).
- `uv run pre-commit run --files openhands/automation/auth.py tests/test_auth.py --show-diff-on-failure` — passed.
- `uv run pytest tests/test_auth.py -q` — unit tests passed, but Docker-backed `TestAuthIntegration` setup failed because this environment has no Docker socket.

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/06d24ec8-1e2f-4ec3-8774-f5c7f7407bca)

### DIFF
--- a/openhands/automation/auth.py
+++ b/openhands/automation/auth.py
@@ -194,18 +194,8 @@ def _credential_cache_key(
     return hashlib.sha256(cache_material.encode()).hexdigest()
 
 
-def _request_has_header(request: Request, name: str) -> bool:
-    try:
-        return name in request.headers
-    except TypeError:
-        return False
-
-
 def _extract_x_org_id(request: Request) -> str | None:
     """Extract and normalize the optional organization scope header."""
-    if not _request_has_header(request, X_ORG_ID_HEADER):
-        return None
-
     header_value = request.headers.get(X_ORG_ID_HEADER, "").strip()
     if not header_value:
         return None
--- a/openhands/automation/auth.py
+++ b/openhands/automation/auth.py
@@ -39,6 +39,7 @@ logger = logging.getLogger("automation.auth")
 # Auth cache - initialized lazily to use config values
 _auth_cache: TTLCache[str, "AuthenticatedUser"] | None = None
 SESSION_COOKIE_NAME = "keycloak_auth"
+X_ORG_ID_HEADER = "X-Org-Id"
 # Keep parity with OpenHands' cookie chunking helper: 8 * 3000 bytes is
 # comfortably above expected session token sizes while staying bounded.
 MAX_SESSION_COOKIE_CHUNKS = 8
@@ -185,9 +186,37 @@ def clear_auth_cache() -> None:
     _get_auth_cache().clear()
 
 
-def _credential_cache_key(credential: str) -> str:
-    """Hash a credential for use as a cache key (never store raw credential)."""
-    return hashlib.sha256(credential.encode()).hexdigest()
+def _credential_cache_key(
+    credential: str, auth_method: AuthMethod, x_org_id: str | None
+) -> str:
+    """Hash auth scope for use as a cache key (never store raw credentials)."""
+    cache_material = "\0".join((auth_method.value, x_org_id or "", credential))
+    return hashlib.sha256(cache_material.encode()).hexdigest()
+
+
+def _request_has_header(request: Request, name: str) -> bool:
+    try:
+        return name in request.headers
+    except TypeError:
+        return False
+
+
+def _extract_x_org_id(request: Request) -> str | None:
+    """Extract and normalize the optional organization scope header."""
+    if not _request_has_header(request, X_ORG_ID_HEADER):
+        return None
+
+    header_value = request.headers.get(X_ORG_ID_HEADER, "").strip()
+    if not header_value:
+        return None
+
+    try:
+        return str(uuid.UUID(header_value))
+    except ValueError as exc:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Invalid X-Org-Id header (must be a UUID)",
+        ) from exc
 
 
 def _is_rate_limited(response: httpx.Response) -> bool:
@@ -446,9 +475,10 @@ async def authenticate_request(
 
     # --- Resolve credential (API key or cookie) ---
     credential, auth_method = _extract_credential(request)
+    x_org_id = _extract_x_org_id(request)
 
     # --- Cache lookup ---
-    cache_key = _credential_cache_key(credential)
+    cache_key = _credential_cache_key(credential, auth_method, x_org_id)
     auth_cache = _get_auth_cache()
     cached_user = auth_cache.get(cache_key)
     if cached_user is not None:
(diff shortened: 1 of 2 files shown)


openhands/automation/auth.py
```diff
@@ -185,9 +186,37 @@ def clear_auth_cache() -> None:
...
-    return hashlib.sha256(credential.encode()).hexdigest()
+def _credential_cache_key(
+    credential: str, auth_method: AuthMethod, x_org_id: str | None
+) -> str:
+    """Hash auth scope for use as a cache key (never store raw credentials)."""
+    cache_material = "\0".join((auth_method.value, x_org_id or "", credential))
+    return hashlib.sha256(cache_material.encode()).hexdigest()
+
+
+def _request_has_header(request: Request, name: str) -> bool:
```
enyst: This `try/except TypeError` exists only so the `MagicMock` fixture works, not because any real `Request` raises it. In production `request.headers` is a Starlette `Headers`, whose `__contains__` returns a real `bool` and whose `get(name, default)` already returns the default for a missing header — so the whole helper is mock-specific machinery leaking into production code.

Cleaner: give the `mock_request` fixture a real `starlette.dataclasses.Headers` (or a real `Request`), then collapse `_extract_x_org_id` to a single `request.headers.get(X_ORG_ID_HEADER, "").strip()`, drop `_request_has_header` entirely, and the `if not header_value` guard becomes redundant. That removes a helper, an exception handler, and a null-check — a net deletion that also makes the tests exercise the same header semantics as prod.

enyst (approved):
**Taste Rating: 🟡 Acceptable** — clean, focused fix for a real bug; one small taste wart.

*Review by OpenHands-DeepSeek-Pro (AI agent), on behalf of the user.*

---

- [openhands/automation/auth.py, line 197] `_request_has_header` — see inline comment. The `try/except TypeError` exists only to accommodate the `MagicMock` test fixture, not any real `Request` behavior (Starlette `Headers.__contains__` never raises `TypeError`, and `get(name, default)` already returns the default for a missing header). Production code shouldn't carry a special case only a mock needs. Cleaner path: give the `mock_request` fixture a real `starlette.dataclasses.Headers` (or `Request`), which collapses `_extract_x_org_id` to a single `request.headers.get(X_ORG_ID_HEADER, "").strip()` and deletes the `_request_has_header` helper plus the now-redundant `if not header_value` guard.

**Correct parts (no action needed):**
- Keying the cache by `(auth_method, normalized_org, credential)` is right, and incidentally fixes a latent collision where an API key and a cookie sharing the same token string would map to the same cache entry.
- `str(uuid.UUID(...))` normalization is good taste — it canonicalizes `urn:uuid:`, braced, and hex forms so the same org can't fragment the cache.
- The new tests exercise the real `authenticate_request` path and assert on forwarded headers and resolved `org_id`/`call_count`, not just mock-call counts. The mocked HTTP boundary is the legitimate seam here.

**[TESTING GAPS]**

- Minor: no assertion that `X-Org-Id` is *absent* from outbound headers when the incoming request omits it. The existing `test_authenticate_valid_api_key` would still pass if someone removed the `if x_org_id:` guard and unconditionally attached the header. One extra `assert "X-Org-Id" not in headers` would lock the guard in.

**[RISK ASSESSMENT]**

- [Overall PR] ⚠️ ✅ LOW

The behavior change is additive (only acts when `X-Org-Id` is present), the cache-key change is a pure in-memory format change, and the only call site of `_credential_cache_key` is this file. The one thing a human should confirm is the security boundary: the fix relies on OpenHands `/api/v1/users/me` rejecting a request when the supplied `X-Org-Id` disagrees with the credential's bound org (`api_key_org_id`). That is the property that stops a user from passing another org's UUID to read its automations. The frontend `client.ts` comment in the OpenHands repo confirms this rejection exists upstream, so I'm treating it as verified rather than blocking — but it is the load-bearing invariant this PR rests on.

**VERDICT:** ✅ Worth merging.

**KEY INSIGHT:** The real contribution isn't the header forwarding — it's recognizing that the auth cache key must encode the full authentication *scope* (method + org), not just the credential, or org scoping silently leaks across requests.

---

> **Improve this review?** If any feedback above seems incorrect or irrelevant to this repository, you can teach the reviewer to do better:
>
> 1. Add a `.agents/skills/custom-codereview-guide.md` file to your branch (or edit it if one already exists) with the `/codereview` trigger and the context the reviewer is missing. See the [customization docs](https://docs.openhands.dev/openhands/usage/use-cases/code-review#customization) for the required frontmatter format.
> 2. Re-request a review — the reviewer reads guidelines from the PR branch, so your changes take effect immediately.
> 3. When your PR is merged, the guideline file goes through normal code review by repository maintainers.
>
> **Resolve with AI?** Install the [iterate skill](https://github.com/OpenHands/extensions/tree/main/skills/iterate) in your agent and run `/iterate` to automatically drive this PR through CI, review, and QA until it's merge-ready.
>
> Was this review helpful? React with 👍 or 👎 to give feedback.

openhands/automation/auth.py
malhotra5: Fixed in 64d5db3: removed the mock-specific header helper, simplified X-Org-Id extraction to `request.headers.get(...).strip()`, updated auth tests to use real Starlette `Headers`, and added coverage that X-Org-Id is absent when omitted.

PR #409, opened 2026-08-28
ci: add issue duplicate checker workflows (parity with sdk/OpenHands)

smolpaws:
HUMAN:
We are starting to see dupes or almost-dupe issues in this repo, we could use the duplicate checker to link and potentially close.

### What

Adds the **issue duplicate checker** workflows to this repo, matching what `OpenHands/software-agent-sdk` and `OpenHands/OpenHands` already run:

- `issue-duplicate-checker.yml` — on new issues (and a daily `cron` sweep + manual `workflow_dispatch`), calls the shared `OpenHands/extensions/plugins/issue-duplicate-checker` action to flag likely duplicates with a `duplicate-candidate` label, and auto-closes aged candidates.
- `remove-duplicate-candidate-label.yml` — clears the `duplicate-candidate` label when a human comments on the issue, so genuine activity un-flags it.

### Why

`automation` has an active issue tracker (75+ open) but no duplicate detection, unlike the sibling repos. This brings it to parity so duplicates get triaged the same way.

### Notes

- The workflow files are the repo-agnostic versions from `OpenHands/OpenHands` (parameterized on `${{ github.repository }}`), copied verbatim so behavior matches the other repos. The action is pinned to the same commit (`fb020c71…`).
- **Secrets:** the duplicate-check job uses `OPENHANDS_API_KEY` (OpenHands Cloud) and `OPENHANDS_BOT_GITHUB_PAT_PUBLIC` (falls back to `github.token`). These are used by the same workflow in the sibling repos; please confirm they're available to this repo at the org level before merging — without `OPENHANDS_API_KEY` the check job will fail.
- No changes to app code; workflows only.

Opened at @enyst's request (parity pass across the main repos).

### DIFF
--- a/.github/workflows/issue-duplicate-checker.yml
+++ b/.github/workflows/issue-duplicate-checker.yml
@@ -1,0 +1,92 @@
+---
+name: Issue Duplicate Check via OpenHands Cloud
+
+on:
+    issues:
+        types: [opened]
+    schedule:
+        - cron: 0 9 * * *
+    workflow_dispatch:
+        inputs:
+            mode:
+                description: Which workflow path to run
+                required: true
+                type: choice
+                options:
+                    - smoke-clone
+                    - issue-check
+                    - auto-close
+                default: smoke-clone
+            issue_number:
+                description: Existing issue number to analyze when mode is issue-check
+                required: false
+                type: number
+            close_after_days:
+                description: Days to wait before auto-closing duplicate candidates in auto-close mode
+                required: false
+                type: number
+                default: 3
+
+permissions:
+    contents: read
+    issues: write
+
+jobs:
+    smoke-clone:
+        if: github.event_name == 'workflow_dispatch' && inputs.mode == 'smoke-clone'
+        runs-on: ubuntu-latest
+        timeout-minutes: 5
+        steps:
+            - name: Clone repository
+              env:
+                  REPOSITORY: ${{ github.repository }}
+              run: |
+                  git clone --depth 1 "https://github.com/${REPOSITORY}.git" /tmp/repo-clone
+                  echo "REPO_HEAD=$(git -C /tmp/repo-clone rev-parse --short HEAD)" >> "$GITHUB_ENV"
+
+            - name: Summarize smoke test
+              run: |
+                  {
+                    echo "## Smoke clone completed"
+                    echo
+                    echo "- Repository cloned to /tmp/repo-clone"
+                    echo "- Repository HEAD: ${REPO_HEAD}"
+                  } >> "$GITHUB_STEP_SUMMARY"
+
+    issue-duplicate-check:
+        if: |
+            github.event_name == 'issues' ||
+            (github.event_name == 'workflow_dispatch' && inputs.mode == 'issue-check' && inputs.issue_number != '')
+        runs-on: ubuntu-latest
+        timeout-minutes: 35
+        concurrency:
+            group: issue-duplicate-check-${{ github.repository }}-${{ github.event.issue.number || inputs.issue_number }}
+            cancel-in-progress: false
+        steps:
+            - name: Run issue duplicate check
+              uses: OpenHands/extensions/plugins/issue-duplicate-checker@fb020c71eb8e7cd2678f0251ab24327d3738da7d
+              with:
+                  mode: issue-check
+                  repository: ${{ github.repository }}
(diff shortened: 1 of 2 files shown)


enyst (commented):
## Code Review — `/codereview`

🟡 **Acceptable** — the files are correct and behave identically to the siblings, but this PR ships a known-broken-by-configuration workflow and hides the failure behind `github.token` fallback.

### What this PR does

Copies two workflow files (verbatim, modulo the `smoke-clone` step) from `OpenHands/OpenHands` / `software-agent-sdk` so `automation` gets issue-duplicate detection. The pinned action commit (`fb020c71…`, 2026-06-02) resolves and its `plugins/issue-duplicate-checker/` tree is unchanged on `main`, so the reference is valid and stable. Good.

### The blocking problem: `issue-check` will fail in this repo

The action's `issue-check` path hard-fails when `OPENHANDS_API_KEY` is absent (action.yml → `Validate duplicate check inputs` → `exit 1`). In this repo there is **no** `OPENHANDS_API_KEY` at repo level — repo secrets are `SONAR_HOST_URL` + `SONAR_TOKEN` only — and the org-secrets endpoint 403s for my token, so I can't confirm an org-level secret exists.

This is not hypothetical: the identical workflow in `OpenHands/OpenHands` is currently red on every `issues` event with exactly this error — `Error: openhands-api-key is required for issue-check mode` (e.g. run 33203722361). So merging as-is means every new issue in `automation` immediately triggers a failing check.

I hold `maintain` (not `admin`) and org secrets aren't visible to me, so **please confirm `OPENHANDS_API_KEY` is org-available before merge**, or this becomes a permanently red check on issue creation.

### Other observations

**1. `github.token` fallback is the same bug in a different key (and a rate-limit trap).** `github-token: ${{ secrets.OPENHANDS_BOT_GITHUB_PAT_PUBLIC || github.token }}`. Neither target repo has `OPENHANDS_BOT_GITHUB_PAT_PUBLIC`, so the fallback always wins here. `github.token` is a per-installation token scoped to *this* repo — fine for posting the notice comment — but the action reuses it as `GITHUB_TOKEN` for raw `urllib` REST calls against `api.github.com`. That puts it in the unauthenticated-user rate bucket (5,000 req/h per user id), and the sibling repos are already burning through it: `software-agent-sdk` had a duplicate-check run die with `API rate limit exceeded for user ID 182006032` (run 33152648707). The workflow has no `github-token` input on the action and `permissions: contents: read` at top level, so the composite action inherits the workflow token and the REST path is effectively unauthenticated. An org-level bot PAT would fix both the identity and the rate bucket.

**2. `duplicate-candidate` label is assumed, not created.** The action's own scripts create/ensure the label (`ensureCandidateLabelOnIssue`, `DUPLICATE_CANDIDATE_LABEL`), so the `remove-duplicate-candidate-label.yml` guard `contains(...labels.*.name, 'duplicate-candidate')` works. But `automation` already has a `duplicate` label; the workflows introduce a separate `duplicate-candidate` label. Cosmetic duplication, but the repo will now carry two near-synonymous labels.

**3. `timeout-minutes` on `remove-duplicate-candidate-label.yml` (10m) and `smoke-clone` (5m) differ from the sdk copy, which has none.** Not a bug — `automation`'s version is actually better — just noting this copy isn't 100% verbatim, which slightly undercuts the "copied verbatim" claim in the PR body. Not worth fixing.

### [RISK ASSESSMENT]
- [Overall PR] ⚠️ Risk Assessment: 🟡 MEDIUM — the file contents are low-risk (two well-formed workflow files), but the change is proposed against a repo that provably lacks the required secret, which means a guaranteed red check on every issue open and a latent rate-limit trap. Recommend: confirm the org secret exists, or add the secret, before merge.

**VERDICT:** ⚠️ **Approve-with-confirmation**: merge only after `OPENHANDS_API_KEY` is confirmed org-available (or the workflow no longer requires it).

**KEY INSIGHT:** The real defect isn't in these two files — it's that the same broken-by-default workflow is being copy-pasted into a fourth repo while its failure mode (missing key, shared rate bucket) is already visible and red in the repos that run it today.

> *This review was generated by an AI agent (OpenHands) on behalf of Engel.*

enyst:
Closing: taking a different route. The shared-action workflow is provably broken here — this repo has no `OPENHANDS_API_KEY` (the action's issue-check path hard-exits without it), and the `github.token` fallback burns the unauthenticated rate bucket. We'll run the duplicate checker as a Cloud Automation instead of per-repo workflows.

PR #417, opened 2026-09-03
feat: add automation draft lifecycle dispatch

### LINKED ISSUE #434: Define automation draft storage and lifecycle semantics
## Problem

Automation setup is now a multi-step flow. Users may enter a name, prompt, trigger, model, repositories, plugin configuration, or raw tarball details over time, and the UI needs to save that incomplete work before it is a valid automation creation request.

The backend also has important existing invariants:

- Schedulers and webhook/event dispatch should only consider real active automations.
- `automation_runs.automation_id` is non-null, so any actual run must point at an `automations.id`.
- The dispatcher expects to load executable configuration from an `Automation` row.
- A user-disabled automation and a draft-backed test automation are both not scheduled, but they mean different things.

Because of those constraints, using only the existing `enabled` boolean is ambiguous. It cannot distinguish an automation the user intentionally disabled from an automation row that exists only because a draft was materialized for a manual test run.

### Desired Behavior

The automation service should support server-backed drafts for incomplete setup flows while preserving clear runtime semantics for real automations and draft-backed test materializations.

Specifically, the design should:

1. Save incomplete setup state without forcing nullable or invalid executable fields into the main `automations` table.
2. Validate draft payloads enough that clients cannot store arbitrary unstructured JSON for a given creation endpoint.
3. Let complete drafts be manually test-dispatched through the existing run and dispatcher pipeline.
4. Keep draft/test materializations out of automatic scheduling and webhook/event matching.
5. Preserve clear semantics between active, inactive, and draft-backed automation rows.

## Constraints

- Drafts may be incomplete, so they cannot always satisfy the final creation schema.
- Test dispatch still creates an `AutomationRun`, and runs currently require an `automation_id`.
- The dispatcher reads tarball, entrypoint, trigger, model, timeout, and related execution fields from `Automation`.
- Backward compatibility with `enabled` should be maintained for existing API callers.

## Proposed direction

Use a separate `automation_drafts` table for saved setup state. Store the draft body as JSON, but validate it with endpoint-specific partial Pydantic schemas for:

- `/v1`
- `/v1/preset/prompt`
- `/v1/preset/plugin`

When a draft becomes complete and the user test-dispatches it, materialize it into a linked `automations` row and mark that row as:

- `enabled = false`
- `lifecycle_status = DRAFT`

This keeps incomplete editing state separate while preserving the existing run and dispatcher contract for manual test runs.

### Store drafts directly in `automations`

This makes draft test runs easy because every draft is already an automation row. However, it requires execution-critical fields to become nullable or overloaded, and it forces scheduler, dispatcher, serialization, and response code paths to handle incomplete automation rows.

### Use only `enabled = false`

This avoids a lifecycle column, but it collapses two different concepts: a real automation disabled by the user and a materialized draft automation used only for test dispatch. That ambiguity leaks into UI, dispatch, git sync, cleanup, analytics, and future state transitions.

### Store drafts as arbitrary JSON only

This is flexible, but it allows endpoint drift and arbitrary payloads. Errors surface too late, and clients lose reliable field-level validation for setup forms.

### Dispatch directly from draft JSON

This avoids materialization, but it conflicts with the current run model and dispatcher contract. Changing those paths would be broader and riskier than creating a linked draft automation row for test dispatch.

### Acceptance Criteria

- [ ] Incomplete drafts can be saved.
- [ ] Endpoint-mismatched or unknown draft fields are rejected early.
- [ ] Complete drafts can be materialized and manually test-dispatched.
- [ ] Materialized draft automations are not picked up by scheduler or event dispatch.
- [ ] Disabled production automations remain distinguishable from draft-backed materialized automations.
- [ ] Existing `enabled` semantics remain backward compatible.

---

This issue was updated by an AI agent (OpenHands) on behalf of the user.

malhotra5:
## Summary

Implements backend support for OSS-10472 automation drafts, lifecycle state, and manual dispatch testing.

- Adds `AutomationState` (`ACTIVE`, `INACTIVE`, `DRAFT`) and `AutomationRun.trigger_source` (`manual`, `cron`, `event`).
- Adds persisted `AutomationDraft` records for incomplete setup UI state.
- Adds `/api/automation/v1/drafts` CRUD endpoints plus `/dispatch` for eligible draft test runs.
- Materializes dispatchable drafts into disabled `DRAFT` automations for manual test runs.
- Keeps the draft row as the source of truth while editing, even after materialization.
- Reuses the same materialized `DRAFT` automation row on each successful draft dispatch, overwriting it from the current valid draft body.
- Allows manual dispatch for inactive/draft automations while keeping scheduler and webhook/event paths limited to active automations.
- Updates git sync serialization/import for lifecycle state.
- Adds `automationDrafts` capability discovery flag.

Fixes https://github.com/OpenHands/automation/issues/434

### Source of truth

`automation_drafts` is the source of truth while the user is in the draft/editing flow. A materialized automation row is an executable projection used for test dispatch, not the canonical draft state.

This matters because a draft can be valid, dispatched once, and then edited back into an incomplete state. In that case, future dispatch must validate the current draft body and fail if incomplete. It must not silently run the older materialized automation config.

### Materialization

When a complete draft is dispatched:

1. The draft body is validated against the endpoint-specific final creation schema.
2. If no materialized row exists, the service creates a linked `automations` row.
3. If a materialized row already exists and is not deleted, the service reuses that same row.
4. The materialized row is overwritten from the current valid draft body.
5. The materialized row is marked `enabled = false` and `lifecycle_status = DRAFT`.
6. A manual run is created against that materialized automation id.
7. The draft row remains in place and records `last_test_run_id`.

The materialized row is updated on successful dispatch, not on every edit.

### Editing after materialization

Editing a draft after it has been materialized updates only the draft entry:

- `automation_drafts.draft_body`
- `automation_drafts.validation_errors`
- `automation_drafts.dispatchable`
- `automation_drafts.updated_at`

The linked materialized automation is left untouched until the user explicitly dispatches the draft again and the draft fully validates.

### Incomplete drafts after a previous dispatch

A draft is allowed to become incomplete again after a previous successful test dispatch.

When that happens:

- the draft remains saved,
- `dispatchable` becomes false,
- validation errors are returned on dispatch,
- the older materialized automation remains only as historical execution context for previous runs,
- the older materialized automation is not treated as current source of truth.

### Delete semantics

Deleting a draft ends the draft flow. The intended behavior is:

- soft-delete the draft row by setting `automation_drafts.deleted_at`,
- also soft-delete the linked materialized automation if it still has `lifecycle_status = DRAFT`,
- preserve historical run rows that point at the materialized automation id,
- prevent the test artifact from appearing in normal automation views or being reused.

### Publish/finalize semantics

Publishing or finalizing a draft should be an explicit product transition, separate from test dispatch. At that point, source of truth moves from `automation_drafts` to a normal `automations` row with `lifecycle_status = ACTIVE` or `INACTIVE`.

## Validation model

Draft bodies are stored as JSON, but not arbitrary unstructured JSON. The service validates and normalizes draft bodies with endpoint-specific partial Pydantic schemas for:

- `/v1`
- `/v1/preset/prompt`
- `/v1/preset/plugin`

This allows incomplete drafts to be saved while still rejecting unknown or endpoint-mismatched fields early.

Final dispatchability is still determined by validating against the real final creation schemas.

### Store drafts directly in `automations`

This makes draft test runs easy because every draft is already an automation row. However, it requires execution-critical fields to become nullable or overloaded, and it forces scheduler, dispatcher, serialization, and response code paths to handle incomplete automation rows.

### Use only `enabled = false`

This avoids a lifecycle column, but it collapses two different concepts: a real automation disabled by the user and a materialized draft automation used only for test dispatch. That ambiguity leaks into UI, dispatch, git sync, cleanup, analytics, and future state transitions.

### Store drafts as arbitrary JSON only

This is flexible, but it allows endpoint drift and arbitrary payloads. Errors surface late, and clients lose reliable field-level validation for setup forms.

### Dispatch directly from draft JSON

This avoids materialization, but it conflicts with the current run model and dispatcher contract: runs require an `automation_id`, and execution config is loaded from `Automation`. Changing that path would be wider and riskier.

## Tests

- `uv run pre-commit run --all-files --show-diff-on-failure`
- `uv run pytest tests/test_draft_router.py tests/test_capabilities_router.py tests/test_dispatcher.py -q` (`72 passed`)
- `uv run pytest tests/ -q --ignore=tests/integration` (`1720 passed, 7 skipped`)
- `uv run alembic heads` (`023 (head)`)
- temporary SQLite `uv run alembic upgrade head`

_This PR description was updated by an AI agent (OpenHands) on behalf of the user._

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/13bfb65f-3c18-4f3a-bf37-848e96a85cef)

### DIFF
--- a/migrations/versions/022_add_automation_drafts.py
+++ b/migrations/versions/022_add_automation_drafts.py
@@ -61,9 +61,7 @@ def upgrade() -> None:
         sa.Column("name", sa.String(length=500), nullable=True),
         sa.Column("draft_body", sa.JSON(), nullable=False),
         sa.Column("validation_errors", sa.JSON(), nullable=True),
-        sa.Column(
-            "dispatchable", sa.Boolean(), nullable=False, server_default="false"
-        ),
+        sa.Column("dispatchable", sa.Boolean(), nullable=False, server_default="false"),
         sa.Column("source_automation_id", sa.Uuid(), nullable=True),
         sa.Column("materialized_automation_id", sa.Uuid(), nullable=True),
         sa.Column("last_test_run_id", sa.Uuid(), nullable=True),
@@ -127,7 +125,7 @@ def upgrade() -> None:
 
     op.execute(
         "COMMENT ON COLUMN automations.lifecycle_status IS "
-        "'First-class lifecycle state: ACTIVE, INACTIVE, or DRAFT.'"
+        "'Automation state: ACTIVE, INACTIVE, or DRAFT.'"
     )
     op.execute(
         "COMMENT ON COLUMN automation_runs.trigger_source IS "
--- a/migrations/versions/022_add_automation_drafts.py
+++ b/migrations/versions/022_add_automation_drafts.py
@@ -1,0 +1,167 @@
+"""Add automation drafts and run trigger source.
+
+Revision ID: 022
+Revises: 021
+Create Date: 2026-09-03
+"""
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+
+revision: str = "022"
+down_revision: str = "021"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def _is_sqlite() -> bool:
+    return op.get_bind().dialect.name == "sqlite"
+
+
+def upgrade() -> None:
+    op.add_column(
+        "automations",
+        sa.Column(
+            "lifecycle_status",
+            sa.String(length=20),
+            nullable=False,
+            server_default="ACTIVE",
+        ),
+    )
+    op.create_index(
+        "ix_automations_lifecycle_status", "automations", ["lifecycle_status"]
+    )
+    op.execute(
+        "UPDATE automations SET lifecycle_status = CASE "
+        "WHEN enabled THEN 'ACTIVE' ELSE 'INACTIVE' END"
+    )
+
+    op.add_column(
+        "automation_runs",
+        sa.Column("trigger_source", sa.String(length=32), nullable=True),
+    )
+    op.create_index(
+        "ix_automation_runs_trigger_source", "automation_runs", ["trigger_source"]
+    )
+    op.create_index(
+        "ix_automation_runs_status_trigger_source",
+        "automation_runs",
+        ["status", "trigger_source"],
+    )
+
+    op.create_table(
+        "automation_drafts",
+        sa.Column("id", sa.Uuid(), nullable=False),
+        sa.Column("user_id", sa.Uuid(), nullable=False),
+        sa.Column("org_id", sa.Uuid(), nullable=False),
+        sa.Column("endpoint", sa.String(length=64), nullable=False),
+        sa.Column("name", sa.String(length=500), nullable=True),
+        sa.Column("draft_body", sa.JSON(), nullable=False),
+        sa.Column("validation_errors", sa.JSON(), nullable=True),
+        sa.Column(
+            "dispatchable", sa.Boolean(), nullable=False, server_default="false"
+        ),
+        sa.Column("source_automation_id", sa.Uuid(), nullable=True),
+        sa.Column("materialized_automation_id", sa.Uuid(), nullable=True),
+        sa.Column("last_test_run_id", sa.Uuid(), nullable=True),
+        sa.Column("deleted_at", sa.DateTime(timezone=True), nullable=True),
+        sa.Column(
+            "created_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("CURRENT_TIMESTAMP"),
+            nullable=False,
+        ),
+        sa.Column(
+            "updated_at",
+            sa.DateTime(timezone=True),
+            server_default=sa.text("CURRENT_TIMESTAMP"),
+            nullable=False,
+        ),
+        sa.ForeignKeyConstraint(
+            ["source_automation_id"], ["automations.id"], ondelete="SET NULL"
+        ),
+        sa.ForeignKeyConstraint(
+            ["materialized_automation_id"], ["automations.id"], ondelete="SET NULL"
+        ),
+        sa.ForeignKeyConstraint(
+            ["last_test_run_id"], ["automation_runs.id"], ondelete="SET NULL"
+        ),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index("ix_automation_drafts_user_id", "automation_drafts", ["user_id"])
+    op.create_index("ix_automation_drafts_org_id", "automation_drafts", ["org_id"])
+    op.create_index(
+        "ix_automation_drafts_deleted_at", "automation_drafts", ["deleted_at"]
+    )
+    op.create_index(
+        "ix_automation_drafts_org_updated_at",
+        "automation_drafts",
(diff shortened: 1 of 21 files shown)


enyst (commented):
Interesting, thank you Rohit!

May I ask a couple of tiny details: is there a GitHub issue for this proposal? 

Also, what I personally try to do on bigger PRs or features is to have the agent make a HTML describing the before and after, or in some way give an “at a glance” with what is changing here. Would you mind to make one, since you know it better? Though I can too if you want 🙏

malhotra5:
Linked the github issue and html breakdown is in the `.pr` folder of this PR

malhotra5:
## Draft behavior matrix

| Action | Draft row | Materialized automation row | Source of truth afterward |
|---|---|---|---|
| Create draft | Create an `automation_drafts` row with the partial endpoint body. Normalize and shape-check the JSON against the endpoint-specific partial schema. | None yet. | `automation_drafts` |
| Edit draft before materialization | Update `draft_body`, recompute `validation_errors`, and recompute `dispatchable` from the final creation schema. | None yet. | `automation_drafts` |
| Dispatch complete draft for the first time | Keep the draft row. Store the new automation id in `materialized_automation_id` and store the run id in `last_test_run_id`. | Create a linked `automations` row with executable fields derived from the draft. Mark it `enabled = false` and `lifecycle_status = DRAFT`. | `automation_drafts` |
| Dispatch incomplete draft | Keep the draft row and return field-level validation errors. Do not dispatch stale previously materialized config. | If an older materialized row exists, leave it as historical execution context only. It is not treated as current config. | `automation_drafts` |
| Edit draft after materialization | Update only the draft entry. The user is editing saved setup state, not directly editing the test-run projection. | Do not eagerly update it. It continues to represent the last complete draft version that was actually test-dispatched. | `automation_drafts` |
| Edit materialized draft into an incomplete state | Allowed. The draft can become non-dispatchable again after a previous successful test dispatch. | Leave the old materialized row untouched until the draft becomes valid and the user explicitly dispatches again. | `automation_drafts` |
| Dispatch complete draft again | Keep the draft row and update `last_test_run_id` to the new run. | Reuse and overwrite the linked `DRAFT` automation row with executable fields derived from the current valid draft body, then create a new manual run. | `automation_drafts` |
| Delete draft | Soft-delete the draft by setting `deleted_at`. | Intended behavior: also soft-delete the linked materialized automation if it still has `lifecycle_status = DRAFT`. This preserves historical run references while removing the test artifact from normal automation views. | None; the draft flow is ended. |
| Delete source automation | Existing draft rows that were seeded from that automation should keep their saved draft body. The source link may be cleared by foreign-key behavior. | Normal automation deletion semantics apply to the source automation. It should not mutate the draft body. | `automation_drafts`, if the draft still exists |
| Publish or finalize draft | Explicit transition separate from test dispatch. The draft may be marked consumed, deleted, or kept for audit depending on the product decision. | Create or convert to a normal automation with `lifecycle_status = ACTIVE` or `INACTIVE`. It is no longer a draft test artifact. | The finalized `automations` row |

Core invariant: editing always targets the draft row. The materialized automation is an executable projection created only for test dispatch. A later edit can make the draft incomplete again, so the older materialized row must never silently become the current source of truth.

_This PR comment was created by an AI agent (OpenHands) on behalf of the user._

PR #415, opened 2026-09-02
feat: split automation permissions into view and manage

tofarr:
## Summary

Implements [OHE-3182](https://linear.app/all-hands-ai/issue/OHE-3182/update-permissions-for-members-to-view-only): Update permissions for members to view-only.

Splits the single `MANAGE_AUTOMATIONS` permission into two:
- `view_automations` — read-only access (list, get, list runs, capabilities, validate, git-sync status, webhook list/get, upload list/get, telemetry consent)
- `manage_automations` — write access (create, update, delete, dispatch, cancel run, preset create, webhook create/update/delete/rotate, upload create/delete, git-sync config/check/sync)

Members now get `view_automations` only (read-only). Admins and owners retain both permissions. The automation creator can still edit their own automation regardless of role — this is a row-level comparison (`automation.user_id == user.user_id`), not a permission string.

### Application code
| File | Change |
|------|--------|
| `auth.py` | Local user gets both `view_automations` + `manage_automations` |
| `router.py` | Read endpoints → `_require_view_automations`; write endpoints on specific automations → view dep + `_assert_can_manage` helper (checks `manage_automations` OR creator); create → manage |
| `webhook_router.py` | list/get → view; create/update/delete/rotate → manage |
| `uploads.py` | list/get → view; create/delete → manage |
| `capabilities_router.py` | get_capabilities + validate → view |
| `preset_router.py` | both preset endpoints → manage |
| `telemetry_router.py` | consent endpoint → view |
| `git_sync/router.py` | GET /status → view; config/check/sync → manage |

### Tests
- `conftest.py`: `mock_authenticated_user` and `mock_readonly_user` updated with new permissions
- `test_router.py`: `TestPermissionEnforcement` now tests non-creator member 403 + creator success (read/write)
- `test_auth.py`: Updated mock response, permission assertions, and local user structure tests
- `test_telemetry.py`, `test_git_sync_router.py`, `test_ab_testing_integration.py`: Updated user fixtures

## Companion PR

This PR depends on the enterprise repo change that defines `VIEW_AUTOMATIONS` in the permission enum and assigns it to the member role.

## Test plan

- [x] All non-Docker unit tests pass (`TestRequirePermission`, `TestLocalApiKeyAuthentication`, `test_ab_testing_integration`)
- [x] Ruff check + format pass on all modified files
- [ ] Full test suite (requires Docker — CI will validate)
- [ ] Manual: verify member can list/get automations but cannot create/update/delete
[...]

PR #427, opened 2026-09-08
feat: restrict automation edits to the creator

hieptl:
### Summary

Implements [OHE-3219](https://linear.app/all-hands-ai/issue/OHE-3219/org-automations-can-only-be-edited-by-the-creator): only the creator may edit an org automation.

Every automation runs as its creator. `dispatcher.py` exports `AUTOMATION_USER_ID = automation.user_id` and `backends/cloud.py` mints the run's API key for that user via `get_api_key_for_automation_run`, so the sandbox uses the creator's git tokens, secrets, and MCP servers. After #415 split the permission into `view_automations` / `manage_automations`, `_assert_can_manage` still let any admin or owner rewrite another member's prompt, tarball, schedule, or model. Product decision on the ticket (and on OHE-3183): the creator edits; admins and owners may turn off, delete, or export.

Because "turn off" is `PATCH /v1/{id}` with `{"enabled": false}` through the same handler as edits, the rule has to be field-aware rather than endpoint-level.

### Behaviour

| Caller | Edit definition (`prompt`, `tarball_path`, `trigger`, `model`, …) | `{"enabled": false}` | `{"enabled": true}` | `DELETE` | Dispatch / cancel |
|---|---|---|---|---|---|
| Creator (any role) | allowed | allowed | allowed | allowed | allowed |
| Admin / owner, not creator | **403** (new) | allowed | **403** (new) | allowed | allowed (unchanged) |
| Member, not creator | 403 (unchanged) | 403 (unchanged) | 403 (unchanged) | 403 (unchanged) | 403 (unchanged) |

### Changes

| File | Change |
|---|---|
| `openhands/automation/router.py` | `update_automation`: after `_assert_can_manage`, reject a non-creator with 403 unless `update_data == {"enabled": False}`. Docstrings on `_assert_can_manage` and `update_automation` updated to describe the narrower rule. |
| `tests/test_router.py` | `TestPermissionEnforcement`: helper `_other_users_automation` plus four tests using the owner client against another user's automation — definition edit → 403, turn off → 200 with `enabled: false`, turn on → 403, turn off combined with a rename → 403. |

Not changed: `delete_automation`, `dispatch_automation`, `cancel_run`, `UpdateAutomationRequest`, preset / webhook / upload / git-sync routers, and the enterprise permission enum. Local mode is unaffected because the local user id is constant, so every local automation's creator is the caller.

### Companion PR

OpenHands/OpenHands (Agent Canvas): hides the "Turn on" switch and menu entry for a disabled automation the caller did not create, so managers are not offered an action this service now rejects. Edit is already local-only there.

### Test plan

Run locally:

```
uv run ruff check openhands/automation/router.py tests/test_router.py
uv run ruff format --check openhands/automation/router.py tests/test_router.py
uv run pytest tests/test_router.py -k "TestPermissionEnforcement or TestUpdateAutomation or TestDeleteAutomation" -q
```

- [x] ruff check and format: clean.
- [x] Targeted suites above: 33 passed (8 in `TestPermissionEnforcement`, 4 of them new).
- [x] Full `tests/test_router.py`: 134 passed, 1 failed. The failure is `TestCreateAutomation::test_create_automation_shares_template_identity_with_presets`, which needs a local S3 endpoint on `localhost:9000` (`EndpointConnectionError`); it exercises the create path and is unrelated to this change. CI has that dependency.
- [ ] Manual against a running service, as an admin on an automation created by another member:
  - `PATCH /api/automation/v1/{id}` with `{"prompt": "x"}` → 403 `Only the automation creator can edit it; admins and owners can only turn it off or delete it`
  - `PATCH` with `{"enabled": false}` → 200, `enabled: false`
  - `PATCH` with `{"enabled": true}` → 403
  - `DELETE /api/automation/v1/{id}` → 204
  - As the creator, any `PATCH` → 200


tofarr (approved):
🍰

PR #429, opened 2026-09-09
feat: scope git sync to organizations so cloud deployments can sync

hieptl:
### Why

Git Sync is not available on cloud backends. The Agent Canvas gate is only the symptom; this service's git sync is single-tenant by design and hard-off outside local mode:

- `is_git_sync_supported()` returned `is_local_mode`, which is never true in the cloud.
- Runtime config was one blob and five status keys in `automation_service_metadata`.
- `automation_git_sync_state` had no `org_id` and a globally unique `slug`.
- Export, backfill, import and path-change queries were unscoped, and imported automations were owned by the fixed local user.
- Concurrency was an in-process `asyncio.Lock`, while production runs several replicas with no shared disk.
- The secret wrapping key fell back to a per-pod key file.

Flipping the frontend alone would have shown a "Disabled" page whose Save wrote a service-wide config visible to every org's admins.

### What changed

- **Data model.** New `AutomationGitSyncOrgConfig` (`automation_git_sync_org_config`): per-org override blob (secrets still Fernet-wrapped), `configured_by_user_id`, `last_synced_commit`, `last_synced_path`, `last_run_at`, `last_error`, `last_error_at`, `sync_started_at`. `AutomationGitSyncState` gains `org_id` and a unique `(org_id, slug)` index.
- **Migration 023.** Creates the table, backfills `org_id` from `automations`, swaps the slug index, and moves a local deployment's legacy service-wide config and status onto the local org's row (a cloud database holding only the blob drops it, since there is no org to attribute it to). Downgrade is best-effort and documented.
- **`git_sync/config_override.py`.** Every helper takes `org_id`; `base_git_sync_settings()` blanks the env repo URL, token and encryption key outside local mode; `find_org_using_repo()` backs the new 409.
- **`git_sync/loop.py`.** Org-scoped queries and workdir, import owner from the org row, bookkeeping on the org row, a DB lease (`_acquire_sync_lease` / `_release_sync_lease`, `lease_is_live`) instead of the module lock, and a loop that enumerates configured orgs and runs the ones that are due. `mark_git_sync_dirty` now runs in every deployment and stamps `org_id`.
- **`git_sync/router.py`.** Every endpoint acts on `user.org_id`; `PUT /config` stamps the configuring user and returns 409 on a repo/branch/path collision; `POST /sync` and `GET /status` read the lease.
- **`git_sync/secret_store.py` and `config.py`.** New `AUTOMATION_GIT_SYNC_SECRET`; key order is that, then `AUTOMATION_KV_SECRET`, then (local mode only) the key file.
- **`app.py`.** The loop starts unconditionally; the env-repo warning explains that env-level repo config is ignored outside local mode.
- **Docs.** `AGENTS.md` Git Sync section rewritten for org scoping; stale docstrings in `config.py`, `schemas.py` and `utils/service_metadata.py` fixed.

### Behavior notes and rollout

- **Cloud prerequisites.** `AUTOMATION_GIT_SYNC_SECRET` must be set (otherwise `PUT /config` with a token returns 503 with an explicit message) and `/workspace` must be writable (the image does not create it). Both are wired in the OpenHands-Cloud chart change.
- **Local mode** keeps its checkout path and env-var configuration; the only visible change is a one-time move of the config row by the migration.
- **Ship order.** This PR and the chart change go first; the Agent Canvas change that unhides the feature must not ship before them.
- **Trust boundary.** Anything pushed to an org's configured repo becomes an automation that runs as the configuring admin. Documented in `AGENTS.md`.

### Testing

Commands run locally (Docker available for the Postgres test container):

```
uv run ruff check openhands migrations tests
uv run ruff format --check openhands migrations tests
uv run pyright openhands/automation/git_sync openhands/automation/models.py openhands/automation/app.py migrations/versions/023_org_scoped_git_sync.py
uv run pytest tests/test_git_sync.py            # 49 passed (SQLite, real bare repo)
uv run pytest tests/test_git_sync_router.py     # 49 passed (PostgreSQL container)
uv run pytest tests/test_db.py                  # 27 passed (alembic upgrade head on SQLite)
uv run pytest tests/ --ignore=tests/integration # 1681 passed, 7 skipped, 1 failed
```

The single failure in the full run is `tests/test_router.py::TestCreateAutomation::test_create_automation_shares_template_identity_with_presets`; it is environmental (the repo `.env` points `FILE_STORE=s3` at a MinIO on `localhost:9000` that is not running) and passes with `FILE_STORE=local`. It does not touch git sync.

Migration 023 was also exercised against PostgreSQL 15 in Docker: `alembic upgrade 022`, seeding of the legacy `git_sync_*` metadata keys, `alembic upgrade head` (org row created, legacy keys removed, `org_id NOT NULL`, new indexes), `alembic downgrade 022` (metadata restored, global slug index recreated, table dropped), `alembic upgrade head` again.

Test changes: the tests that codified local-only behavior were updated (`TestGitSyncConfigOutsideLocalMode`, the in-progress/lease tests, every `run_sync_cycle` call now names the org), and new coverage was added for org isolation of exports, import ownership, per-org status/config, the same-repo 409 and the different-path allowance, the expired-lease takeover, and the 503 without a wrapping key outside local mode.

### Follow-ups

- Ops: provision `automation-git-sync-secret` (SOPS) for production and staging and add `gitSyncSecretFromSecret` to the per-environment saas-deploy values.
- A heartbeat refresh of the lease for very long clones (today a cycle that outlives the TTL can lose a push to a non-fast-forward error, which self-heals next cycle).


tofarr (changes requested):
## Code Review: Scope git sync to organizations

Overall this is a high-quality, well-reasoned change. The data-model move from service-wide metadata to a per-org `automation_git_sync_org_config` table is the right call, the cross-replica DB lease is a clean replacement for the in-process `asyncio.Lock`, the secret-wrapping key precedence is sensible, and the migration's handling of the local-org attribution edge case is carefully thought through. Docstrings are excellent. I ran `ruff check`, `ruff format --check`, `pyright` (all clean), `tests/test_db.py` (27 passed) and `tests/test_git_sync.py`.

I'm requesting changes for one real bug and a couple of smaller items.

---

### 🔴 Blocking: lease comparison raises `TypeError` on SQLite (naive vs aware datetime)

`tests/test_git_sync.py::TestRunSyncCycle::test_concurrent_cycles_are_serialized` fails on SQLite with:

```
TypeError: can't compare offset-naive and offset-aware datetimes
sqlalchemy/orm/evaluator.py:264: TypeError
```

Root cause is in `_acquire_sync_lease` (`loop.py`):

```python
or_(
    AutomationGitSyncOrgConfig.sync_started_at.is_(None),
    AutomationGitSyncOrgConfig.sync_started_at < started_at - ttl,   # <- aware
)
```

`started_at = utcnow()` is tz-aware. SQLAlchemy's default update `synchronize_session="evaluate"` evaluates the WHERE clause **in Python** against the in-memory ORM object. SQLite ignores `DateTime(timezone=True)` and returns `sync_started_at` as a **naive** datetime, so once a row has a non-`None` `sync_started_at`, the `is_(None)` short-circuit no longer applies and `naive < aware` raises. PostgreSQL (`timestamptz`) normalizes on write so it doesn't reproduce there, which is presumably why it wasn't caught.

This is the exact scenario the lease exists for (loop + manual `/sync`, or two replicas), and it affects local mode — a supported deployment — where SQLite is the default backend. Note `lease_is_live` already calls `ensure_utc()` for the same reason; the acquire/release UPDATEs need the same treatment.

Suggested fix: normalize to naive UTC for the comparison, or use `ensure_utc` consistently. The cleanest option is to pass `started_at.replace(tzinfo=None)` (UTC-naive) to both the acquire and release comparisons, since SQLite stores naive and PG ignores the offset on `timestamptz` anyway. Alternatively, set `synchronize_session=False` on both UPDATEs and rely on `rowcount` — that sidesteps Python-side evaluation entirely and is arguably more correct for a conditional UPDATE whose matching you want the DB, not the ORM, to decide.

(Note: in this sandbox `tests/test_git_sync.py` shows two failures. `test_inactive_without_local_mode` is environmental — `AUTOMATION_AGENT_SERVER_URL` is ambient here so `is_local_mode` is true and the assertion flips; not a PR bug. The concurrency failure reproduces in a clean env with `AUTOMATION_AGENT_SERVER_URL` unset.)

---

### 🟡 `app.py`: version downgraded `1.11.1` -> `1.11.0`

```diff
-        version="1.11.1",  # x-release-please-version
+        version="1.11.0",  # x-release-please-version
```

Looks like a merge/rebase artifact — `main` is at `1.11.1`. release-please owns this line, so a downgrade here will either conflict on merge or produce a bogus release. Should be restored to `1.11.1` (or whatever the target is at merge time).

---

### 🟡 Unrelated `AGENTS.md` edits

The diff includes changes to the "PR Artifacts" workflow section (fork PR `.pr/` cleanup behavior) that have nothing to do with org-scoping git sync. If intentional, fine; if not, consider dropping them so the PR stays focused.

---

### 🟡 TOCTOU in `PUT /config` repo collision check

`find_org_using_repo` runs, then `apply_git_sync_config_override` writes — two orgs configuring the same repo+branch+path concurrently can both pass the check and both write. The window is small, but the consequence is exactly what the 409 prevents (two orgs importing each other's automations), a real trust/operational hazard in cloud. Worth a `SELECT ... FOR UPDATE` around check+apply, or a constraint. Not strictly blocking given the narrow window and admin-only access, but flagging it.

---

### 🟢 Minor / nits

- **`_import_owner` trust assumption**: automations imported from git run as `configured_by_user_id`. If that admin is later removed from the org or demoted, their automations keep running as them and their API key keeps getting minted. The docstring notes the trust boundary for *pushed* content, but there's no re-validation that the configured-by user is still an active org member at cycle time. Worth a follow-up.
- **`_handle_path_change`** leaves the old directory in the repo and only logs — documented and conservative. Good.
- **`find_org_using_repo`** does a full table scan per org config row. Justified by the comment for now; if org counts grow, an expression index or denormalized identity column would help. Not blocking.
- **`_resolve_workdir`** in cloud uses `{workspace_base}/git-sync/{org_id}` on a shared pod disk; the lease guarantees a single cycle per org, so no concurrent writes to the same path. Good (and called out in AGENTS.md).
- **Migration downgrade** is best-effort and documented — re-creating the global unique slug index fails if two orgs share a slug. Acceptable for a downgrade.

---

### Summary

Design is sound and implementation is careful. The one change I'd ask for before merge is the lease `TypeError` fix (it breaks the feature's headline concurrency guarantee on the default local backend). The version downgrade and unrelated AGENTS.md edit should be cleaned up. The TOCTOU and stale-owner items are reasonable as follow-ups.

_This review was created by an AI agent (OpenHands) on behalf of the reviewer._

hieptl:
Hello @tofarr,

I have updated the code based on your feedback.

Thank you very much for taking the time to review and provide your valuable feedback! 🙏

tofarr (approved):
🍰

PR #438, opened 2026-09-10
feat: add lifecycle_status enum and trigger_source for automation runs

malhotra5:
## Summary

This is **PR 1 of 2** in a stacked PR series that splits [PR #417](https://github.com/OpenHands/automation/pull/417) (automation draft lifecycle dispatch) into smaller, independently reviewable pieces.

### PR 1 (this PR): lifecycle_status + trigger_source foundation

Introduces the data model and plumbing for first-class automation lifecycle states and run trigger sources, without yet adding the draft-specific endpoints or synthetic-payload test dispatch (those come in PR 2).

### PR 2 (follow-up): draft lifecycle endpoints + event test dispatch

Will add the drafts table, draft schemas/endpoints, event automation test dispatch in draft mode, and synthetic payload support for manual test runs. **Stacked on top of this PR.**

---

### 1. Deprecates `enabled` flag in favor of `lifecycle_status` enum

- Adds `AutomationState` enum (`ACTIVE`/`INACTIVE`/`DRAFT`) to `models.py`
- Adds `lifecycle_status` column on `Automation` (indexed, backfilled from `enabled` in migration)
- Keeps `enabled` for backwards compatibility — only `ACTIVE` rows have `enabled=True`
- All places that previously set `enabled=False` now also set `lifecycle_status=INACTIVE`:
  - `router.py`: create, update, delete
  - `scheduler.py`: disable invalid cron
  - `utils/run.py`: `disable_automation`
  - `utils/unhealthy.py`: `_apply_disable`
  - `git_sync/loop.py`: soft-delete on import

### 2. Adds `trigger_source` to `AutomationRun`

- New `trigger_source` column (`manual`/`cron`/`event`/`null` for legacy)
- Set to `"manual"` on dispatch endpoint, `"cron"` in scheduler, `"event"` in webhook handler
- **Manual test dispatches now work for both cron AND event automations**, even when inactive
- `dispatcher.py`: `_poll_pending_runs` now dispatches manual runs regardless of automation enabled/state
- `skip_pending_runs_for_disabled_automation` excludes manual runs by default (`include_manual=False`); delete uses `include_manual=True`

### 3. Schema-level enforcement

- `AutomationState` StrEnum in `schemas.py`
- `normalize_automation_state_enabled()` model_validator keeps `enabled` and `lifecycle_status` consistent
- `lifecycle_status` field added to `CreateAutomationRequest`, `UpdateAutomationRequest`, `AutomationResponse`, and both preset request models
- `trigger_source` added to `AutomationRunResponse`

### 4. Scheduler/webhook filtering

- `_fetch_enabled_automations` (scheduler) and `get_event_automations`/`get_requested_event_types` (webhook) now filter on `lifecycle_status == ACTIVE` in addition to `enabled`
- Draft automations are never triggered automatically

### 5. Migration

- `023_add_lifecycle_status_and_trigger_source.py`: adds columns, indexes, and backfills `lifecycle_status` from `enabled`

---

## Test plan

- [x] All 1714 existing unit tests pass (`uv run pytest tests/ -q --ignore=tests/integration`)
- [x] New test: `test_dispatches_manual_run_for_disabled_automation` — manual run dispatched even when automation is INACTIVE
- [x] New test: `test_dispatch_disabled_automation_creates_manual_run` — dispatch endpoint creates manual run instead of returning 409
- [x] New test: `test_poll_excludes_draft_even_if_enabled_flag_is_true` — draft automations never scheduled
- [x] New test: `test_excludes_draft_automations` — draft automations excluded from event matching
- [x] Removed tests for the old "only creator can edit" restriction (relaxed in favor of draft workflow)

---

*This PR was created by an AI agent (OpenHands) on behalf of the user.*

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/204ca120-d6ca-470c-9ab4-ecf4265951a2)

### DIFF
--- a/openhands/automation/git_sync/loop.py
+++ b/openhands/automation/git_sync/loop.py
@@ -54,6 +54,7 @@ from openhands.automation.git_sync.serializer import (
 from openhands.automation.models import (
     Automation,
     AutomationGitSyncState,
+    AutomationState,
     TarballUpload,
     UploadStatus,
 )
@@ -474,6 +475,14 @@ async def _validate_and_resolve_fields(
         session, fields, deserialized, slug, existing, pending_storage_deletes
     )
 
+    enabled = True if fields.get("enabled") is None else bool(fields["enabled"])
+    lifecycle_status = fields.get("lifecycle_status")
+    if lifecycle_status == AutomationState.DRAFT.value:
+        lifecycle = AutomationState.DRAFT
+        enabled = False
+    else:
+        lifecycle = AutomationState.ACTIVE if enabled else AutomationState.INACTIVE
+
     return {
         "name": name,
         "model": fields.get("model"),
@@ -482,10 +491,8 @@ async def _validate_and_resolve_fields(
         "setup_script_path": setup_script_path,
         "timeout": timeout,
         "keep_alive": fields.get("keep_alive"),
-        # `dict.get`'s default only applies when the key is absent. A hand edit
-        # leaving "enabled:" empty is valid YAML parsing to None, and
-        # bool(None) would silently disable a live automation on import.
-        "enabled": True if fields.get("enabled") is None else bool(fields["enabled"]),
+        "enabled": enabled,
+        "lifecycle_status": lifecycle,
         "prompt": fields.get("prompt"),
         "preset_metadata": fields.get("preset_metadata"),
         "tarball_path": tarball_path,
@@ -718,6 +725,7 @@ async def _import_from_git(
         automation = await session.get(Automation, state.automation_id)
         if automation is not None and automation.deleted_at is None:
             automation.enabled = False
+            automation.lifecycle_status = AutomationState.INACTIVE
             automation.deleted_at = utcnow()
             result.deleted_in_db += 1
             logger.info("Soft-deleted automation %s (removed from git)", automation.id)
--- a/openhands/automation/git_sync/loop.py
+++ b/openhands/automation/git_sync/loop.py
@@ -476,12 +476,14 @@ async def _validate_and_resolve_fields(
     )
 
     enabled = True if fields.get("enabled") is None else bool(fields["enabled"])
-    lifecycle_status = fields.get("lifecycle_status")
-    if lifecycle_status == AutomationState.DRAFT.value:
-        lifecycle = AutomationState.DRAFT
+    state = fields.get("state")
+    if state == AutomationState.DRAFT.value:
+        automation_state = AutomationState.DRAFT
         enabled = False
     else:
-        lifecycle = AutomationState.ACTIVE if enabled else AutomationState.INACTIVE
+        automation_state = (
+            AutomationState.ACTIVE if enabled else AutomationState.INACTIVE
+        )
 
     return {
         "name": name,
@@ -492,7 +494,7 @@ async def _validate_and_resolve_fields(
         "timeout": timeout,
         "keep_alive": fields.get("keep_alive"),
         "enabled": enabled,
-        "lifecycle_status": lifecycle,
+        "state": automation_state,
         "prompt": fields.get("prompt"),
         "preset_metadata": fields.get("preset_metadata"),
         "tarball_path": tarball_path,
@@ -725,7 +727,7 @@ async def _import_from_git(
         automation = await session.get(Automation, state.automation_id)
         if automation is not None and automation.deleted_at is None:
             automation.enabled = False
-            automation.lifecycle_status = AutomationState.INACTIVE
+            automation.state = AutomationState.INACTIVE
             automation.deleted_at = utcnow()
             result.deleted_in_db += 1
             logger.info("Soft-deleted automation %s (removed from git)", automation.id)
--- a/openhands/automation/models.py
+++ b/openhands/automation/models.py
@@ -102,12 +102,12 @@ class Automation(Base):
     # means the automation service owns explicit cleanup.
     keep_alive: Mapped[bool | None] = mapped_column(default=None, nullable=True)
 
-    # Deprecated: use lifecycle_status instead. Kept for backwards
+    # Deprecated: use state instead. Kept for backwards
     # compatibility; only ACTIVE rows have enabled=True. Will be removed in a
     # future release.
     enabled: Mapped[bool] = mapped_column(default=True, nullable=False, index=True)
 
-    lifecycle_status: Mapped[AutomationState] = mapped_column(
+    state: Mapped[AutomationState] = mapped_column(
         Enum(AutomationState, native_enum=False, length=20),
         nullable=False,
         default=AutomationState.ACTIVE,
--- a/openhands/automation/models.py
+++ b/openhands/automation/models.py
@@ -8,6 +8,7 @@ from typing import Any
 from sqlalchemy import (
     JSON,
     BigInteger,
+    Boolean,
     DateTime,
     Enum,
     Float,
@@ -46,6 +47,14 @@ class AutomationRunStatus(enum.Enum):
     FAILED = "FAILED"
     CANCELLED = "CANCELLED"
     SKIPPED = "SKIPPED"
+
+
+class AutomationState(enum.Enum):
+    """State of an automation definition."""
+
+    ACTIVE = "ACTIVE"
+    INACTIVE = "INACTIVE"
+    DRAFT = "DRAFT"
 
 
 class Automation(Base):
@@ -94,8 +103,19 @@ class Automation(Base):
     # means the automation service owns explicit cleanup.
     keep_alive: Mapped[bool | None] = mapped_column(default=None, nullable=True)
 
-    # Whether the automation is enabled (can be triggered)
+    # Whether the automation is enabled (can be triggered automatically).
+    # Kept for backwards compatibility; lifecycle_status stores the
+    # active/inactive/draft automation state. Only ACTIVE rows should have
+    # enabled=True.
     enabled: Mapped[bool] = mapped_column(default=True, nullable=False, index=True)
+
+    lifecycle_status: Mapped[AutomationState] = mapped_column(
+        Enum(AutomationState, native_enum=False, length=20),
+        nullable=False,
+        default=AutomationState.ACTIVE,
+        server_default=AutomationState.ACTIVE.value,
+        index=True,
+    )
 
     # Current disabled-state metadata. AutomationDisableEvent keeps history.
     disabled_reason: Mapped[str | None] = mapped_column(Text, nullable=True)
@@ -218,6 +238,11 @@ class AutomationRun(Base):
     # local mode). Set immediately after `_start_bash` returns.
     bash_command_id: Mapped[str | None] = mapped_column(String(64), nullable=True)
 
+    # How this run was created: manual, cron, event, or null for legacy rows.
+    trigger_source: Mapped[str | None] = mapped_column(
+        String(32), nullable=True, index=True
+    )
+
     # Event payload for event-triggered runs (JSON)
     # Contains the webhook payload that triggered this run.
     # For GitHub events: model_dump() of the parsed Pydantic event
@@ -258,6 +283,7 @@ class AutomationRun(Base):
         Index("ix_automation_runs_status", "status"),
         Index("ix_automation_runs_status_created_at", "status", "created_at"),
         Index("ix_automation_runs_status_timeout_at", "status", "timeout_at"),
+        Index("ix_automation_runs_status_trigger_source", "status", "trigger_source"),
         # Partial: only live subjects are ever looked up, and only
         # `continue_conversation` runs set one.
         Index(
--- a/openhands/automation/models.py
+++ b/openhands/automation/models.py
@@ -103,10 +103,9 @@ class Automation(Base):
     # means the automation service owns explicit cleanup.
     keep_alive: Mapped[bool | None] = mapped_column(default=None, nullable=True)
 
-    # Whether the automation is enabled (can be triggered automatically).
-    # Kept for backwards compatibility; lifecycle_status stores the
-    # active/inactive/draft automation state. Only ACTIVE rows should have
-    # enabled=True.
+    # Deprecated: use lifecycle_status instead. Kept for backwards
+    # compatibility; only ACTIVE rows have enabled=True. Will be removed in a
+    # future release.
     enabled: Mapped[bool] = mapped_column(default=True, nullable=False, index=True)
 
     lifecycle_status: Mapped[AutomationState] = mapped_column(
--- a/openhands/automation/models.py
+++ b/openhands/automation/models.py
@@ -8,7 +8,6 @@ from typing import Any
 from sqlalchemy import (
     JSON,
     BigInteger,
-    Boolean,
     DateTime,
     Enum,
     Float,
--- a/openhands/automation/router.py
+++ b/openhands/automation/router.py
@@ -4,7 +4,7 @@ import asyncio
 import logging
 import re
 import uuid
-from typing import Any
+from typing import Any, Final
 
 from fastapi import (
     APIRouter,
@@ -39,7 +39,6 @@ from openhands.automation.schemas import (
     AutomationResponse,
     AutomationRunListResponse,
     AutomationRunResponse,
-    AutomationState,
     CreateAutomationRequest,
     RunCompleteRequest,
     RunPhaseRequest,
@@ -69,6 +68,10 @@ from openhands.automation.utils.run_status_detail import (
     run_status_detail_from_callback_error,
 )
 from openhands.automation.utils.sandbox import cleanup_sandbox
+from openhands.automation.utils.state import (
+    automation_state_enabled,
+    model_automation_state,
+)
 from openhands.automation.utils.tarball_validation import (
     is_http_url,
     parse_internal_upload_id,
@@ -87,18 +90,7 @@ logger = logging.getLogger(__name__)
 router = APIRouter(prefix="/v1", tags=["Automations"])
 
 
-def _model_automation_state(
-    state: AutomationState | str | None, enabled: bool
-) -> ModelAutomationState:
-    if state is not None:
-        return ModelAutomationState(str(state))
-    return ModelAutomationState.ACTIVE if enabled else ModelAutomationState.INACTIVE
-
-
-def _automation_state_enabled(
-    state: ModelAutomationState,
-) -> bool:
-    return state == ModelAutomationState.ACTIVE
+_STATE_ONLY_UPDATE_FIELDS: Final[frozenset[str]] = frozenset({"state", "enabled"})
 
 
 _require_view_automations = require_permission("view_automations")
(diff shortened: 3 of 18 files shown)


openhands/automation/router.py
```diff
@@ -248,30 +269,27 @@ async def update_automation(
     # already-deleted object.
     session: AsyncSession = Depends(get_session, scope="function"),
 ) -> AutomationResponse:
-    """Partially update an automation.
-
-    Only the creator may edit the definition. Admins and owners may set
-    ``enabled`` to ``False`` (turn it off) but nothing else.
-    """
+    """Partially update an automation."""
```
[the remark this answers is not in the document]
malhotra5: Fixed in `aa6a0de`: non-creators with `manage_automations` can now update only `state`/`enabled`; changes to definition fields such as prompt, tarball, entrypoint, trigger, model, or timeout now return 403. Added coverage for both the rejected definition edit and allowed state update.

openhands/automation/schemas.py
```diff
@@ -335,6 +336,51 @@ class RunStatus(StrEnum):
...
 
+class AutomationState(StrEnum):
+    """State of an automation definition."""
+
+    ACTIVE = "ACTIVE"
+    INACTIVE = "INACTIVE"
+    DRAFT = "DRAFT"
+
+
+DraftEndpoint = Literal["/v1", "/v1/preset/prompt", "/v1/preset/plugin"]
```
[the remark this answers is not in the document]
malhotra5: Fixed in `aa6a0de`: removed the unused `DraftEndpoint` alias from this base PR. It is introduced only in the stacked draft PR where the draft schemas actually use it.

openhands/automation/router.py
```diff
@@ -84,6 +86,21 @@
 
 router = APIRouter(prefix="/v1", tags=["Automations"])
 
+
+def _model_automation_state(
```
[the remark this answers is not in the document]
malhotra5: Fixed in `aa6a0de`: extracted the duplicated state helpers to `openhands/automation/utils/state.py` and updated both `router.py` and `preset_router.py` to use the shared functions.

enyst:
@malhotra5 I had Astra make a pass, and the results were:

> I would fix these before merging the stack. Reviewed [#438](https://github.com/OpenHands/automation/pull/438)⁠￼ at aa6a0de and [#439](https://github.com/OpenHands/automation/pull/439)⁠￼ at e33a93b.
> 
> 1. P1 — #439 permits replacement code to execute under another user’s identity.
>     After Alice test-dispatches her draft, another admin can edit it and dispatch again. [Reusing the materialized automation](https://github.com/OpenHands/automation/blob/e33a93b4cb3292977017fa4e659cc5369e884ea0/openhands/automation/draft_router.py#L440-L465)⁠￼ preserves Alice’s user_id while replacing its code. I reproduced the normal edit endpoint returning 403, the draft route succeeding, and the execution-key request still targeting Alice. Enforce creator ownership before modifying or reusing that artifact.
> 2. P1 — Both migration IDs collide with current main.
>     The stack introduces revisions [023](https://github.com/OpenHands/automation/blob/aa6a0dea23ab411c2df910bebc7825e371c50fcc/migrations/versions/023_add_state_and_trigger_source.py#L14-L15)⁠￼ and [024](https://github.com/OpenHands/automation/blob/e33a93b4cb3292977017fa4e659cc5369e884ea0/migrations/versions/024_add_automation_drafts.py#L14-L15)⁠￼, which main already uses for other migrations. Combining the files produces duplicate revisions, and Alembic rejects head. Renumber and reconnect both migrations after updating the stack.
> 3. P2 — The normal API bypasses draft validation.
>     Test-dispatch a complete draft, then edit it into an incomplete one. The draft endpoint correctly returns 422, but the [ordinary dispatch endpoint](https://github.com/OpenHands/automation/blob/e33a93b4cb3292977017fa4e659cc5369e884ea0/openhands/automation/router.py#L520-L528)⁠￼ still queues its old configuration with 201. A normal PATCH also activates that stale configuration. Draft artifacts need validation against the current draft before dispatch or finalization.
> 4. P2 — A second test dispatch changes what an already queued test executes.
>     Dispatch version A, edit to B, and dispatch again before the dispatcher polls: both runs load B. The [shared automation row is overwritten](https://github.com/OpenHands/automation/blob/e33a93b4cb3292977017fa4e659cc5369e884ea0/openhands/automation/draft_router.py#L463-L465)⁠￼. Preserve each run’s execution configuration, or prevent reuse while an earlier run remains pending.
> 5. P2 — Git sync ignores ACTIVE and INACTIVE state values.
>     [Import derives these states from ](https://github.com/OpenHands/automation/blob/aa6a0dea23ab411c2df910bebc7825e371c50fcc/openhands/automation/git_sync/loop.py#L478-L486)⁠￼[enabled](https://github.com/OpenHands/automation/blob/aa6a0dea23ab411c2df910bebc7825e371c50fcc/openhands/automation/git_sync/loop.py#L478-L486)⁠￼. I imported state: INACTIVE without enabled; it became ACTIVE and was selected by the scheduler. Changing an exported draft’s state to ACTIVE can conversely leave it inactive. Use consistent state normalization across Git and API inputs.
> 
> One additional behavior needs a decision: two failed draft tests plus the first production failure trigger automatic disabling after activation; test history remains in the failure count.
> 
> Targeted reproductions exposed the issues above.

They look plausible to me, but I think maybe some are actually reasonable behaviors: for one, point 4 sounds like pretty good behavior to me. If I dispatched, and then was like, actually let me tweak this prompt, then I’d expect any later run to get the tweak. So new stuff applies.
Point 3 seems like just that: I’d want the old configuration to no longer apply, only the new one. So I’d personally fix 3 and not fix 4.
Not sure about Point 1.

I can ask the agent to post its repro tests, so we see them red?

malhotra5:
@enyst update on the review pass:

Addressed in the stack:

1. Point 2 — migration revision collisions
   - #438 now uses migration revision `025` for lifecycle state / trigger source.
   - #439 now uses migration revision `026` for automation drafts, with `down_revision = "025"`.
   - Both branches were updated against current `main`/base, and local merge checks are clean.

2. Point 3 — normal API bypassing current draft validation
   - Addressed in #439.
   - Normal dispatch and normal activation now guard materialized draft artifacts by checking the current draft record instead of blindly using the stale materialized automation row.
   - If the current draft is invalid, normal dispatch/activation returns the same 422 validation-error shape as draft dispatch.
   - If there is no current draft context, normal API use of draft artifacts is rejected so callers must go through the draft API.

3. Point 5 — Git sync ignoring `ACTIVE` / `INACTIVE` state
   - Addressed in #438.
   - Git import now uses the same state normalization rules as API inputs.
   - Legacy `enabled`-only YAML still works, consistent `state` + `enabled` YAML works, and contradictory pairs are rejected.
   - This covers the `state: INACTIVE` without `enabled` case so it no longer becomes scheduled as active.

Intentionally not addressed:

4. Point 4 — second test dispatch changes what a not-yet-started queued test executes
   - We are keeping this behavior. The desired behavior is that edits made before a queued draft test is actually dispatched can affect that not-yet-started run.
   - I removed/left out regression coverage that asserted the opposite, so CI is not protecting against this desired behavior.

Not addressed in this pass:

5. Point 1 — cross-admin replacement under the original materialized automation owner
   - I did not change this yet because it needed a product/security decision. The current pass focused on the points we were sure about: 2, 3, and 5, while explicitly preserving point 4 behavior.

Current CI status after the latest pushes:
- #438: checks passing at `e859927`.
- #439: checks passing at `9def6b3` after rerunning a transient MinIO image-pull failure.

enyst (dismissed):
🛠️ **REQUEST_CHANGES** · Risk: **HIGH** · `e8599274cb11`

**OpenHands-Astra, helping Engel Nyst (@enyst).** AI review.

Please fix the state-handling defects before relying on this lifecycle foundation.

- **[P2] Validate booleans before comparing state:** Requests combining state with string-valued enabled flags can reject consistent input or silently activate contradictory input because validation uses Python truthiness. [evidence](<https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/schemas.py#L362-L373>)

- **[P2] Preserve the reactivation history boundary:** Existing automations cycled ACTIVE→DRAFT→ACTIVE retain pre-draft failures, allowing the next failure to disable them before reaching a fresh failure threshold. [evidence](<https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/router.py#L313-L339>)

- **[P2] Reject invalid state before uploading:** Git imports containing new or changed tarball content upload it before rejecting invalid state metadata, leaving storage objects without database records. [evidence](<https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/git_sync/loop.py#L508-L519>)

**Limits:** Source-only assessment: tests, database migrations, and live dispatch were not executed; the supplied test-success claim is unverified.; Human confirmation is needed for allowing managers to activate and manually execute another creator's drafts under that creator's identity.

🤖 Security catch credited, material misses confirmed; policy unknown. The bot celebrated its own dismissed approval.

<details>
<summary>Receipts and review scores</summary>

**Finding receipts**

- **[P2] Validate booleans before comparing state:** https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/schemas.py#L362-L373 · https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/schemas.py#L609-L621 · https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/preset_router.py#L230-L239 · https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/preset_router.py#L743-L748

- **[P2] Preserve the reactivation history boundary:** https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/router.py#L313-L339 · https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/utils/unhealthy.py#L104-L137 · https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/router.py#L285-L293 · https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/tests/test_unhealthy_automations.py#L502-L535

- **[P2] Reject invalid state before uploading:** https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/git_sync/loop.py#L508-L519 · https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/git_sync/loop.py#L446-L466 · https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/git_sync/loop.py#L337-L372 · https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/openhands/automation/git_sync/loop.py#L719-L762 · https://github.com/OpenHands/automation/blob/e8599274cb11726d528dbeaa350efe73b8ceb094/tests/test_git_sync_state.py#L44-L70

**Reviews (5):** 2 major misses · 2 supported · 1 unscored

**Reporting:** ❌ 1 failed · 3 supported · 1 unknown.

Code: 0 material miss · 1 minor · 2 defensible · 3 supported · unscored: insufficient evidence. Policy and reporting are judged separately.

| Review @ commit | Code | Policy | Report | Reason / receipt |
| --- | --- | --- | --- | --- |
| [5181014347](#pullrequestreview-5181014347) @ `4fde9ff5db6b` | 0 | UNKNOWN | PASS | Caught the permission regression, but missed Git imports activating explicitly inactive automations. Limits: Historical instruction delivery and runtime validation records are unavailable; the technical misses are established from source, not executed reproductions. [evidence](<https://github.com/OpenHands/automation/blob/4fde9ff5db6b696edd32ed6d529fe70849c304d3/openhands/automation/git_sync/loop.py#L478-L498>) |
| [5181315428](#pullrequestreview-5181315428) @ `aa6a0dea23ab` | 3 | UNKNOWN | PASS | The claimed definition-edit restriction and both added regression tests are present. Limits: Assessment covers the reported permission repair only; tests were not executed and historical posting instructions are unavailable. [evidence](<https://github.com/OpenHands/automation/blob/aa6a0dea23ab411c2df910bebc7825e371c50fcc/openhands/automation/router.py#L123-L135>) |
| [5181315959](#pullrequestreview-5181315959) @ `aa6a0dea23ab` | unscored | UNKNOWN | UNKNOWN | Removal is confirmed, but unavailable stacked source prevents scoring the complete claim. Limits: Supply the stacked PR's historical source to verify its alias usage; the base-PR removal is confirmed.; Instructions applicable to this reviewer at submission time are unavailable. [evidence](<https://github.com/OpenHands/automation/pull/438#discussion_r3991465097>) |
| [5181316276](#pullrequestreview-5181316276) @ `aa6a0dea23ab` | 3 | UNKNOWN | PASS | Both routers now use the shared helpers, confirming the reported extraction. Limits: Assessment is limited to the reported extraction; no runtime validation or applicable historical posting instructions are available. [evidence](<https://github.com/OpenHands/automation/blob/aa6a0dea23ab411c2df910bebc7825e371c50fcc/openhands/automation/utils/state.py#L13-L24>) |
| [5181507684](#pullrequestreview-5181507684) @ `aa6a0dea23ab` | 0 | UNKNOWN | FAIL | Missed Git activation and boolean-validation defects despite correctly recognizing the permission repair. Limits: The governing historical instructions and claimed programmatic-validation transcript are unavailable; neither is treated as verified.; The approval/dismissal contradiction is established by supplied events; source-only technical scoring excludes later code and unavailable stacked behavior. [evidence](<https://github.com/OpenHands/automation/blob/aa6a0dea23ab411c2df910bebc7825e371c50fcc/openhands/automation/git_sync/loop.py#L478-L500>) |

</details>

_Generated by OpenHands AI using openai/gpt-6-astra; reviewed openhands/automation#438 at `e8599274cb11726d528dbeaa350efe73b8ceb094`._

migrations/versions/026_add_state_and_trigger_source.py
```diff
@@ -0,0 +1,60 @@
+"""Add automation state and run trigger_source.
```
VascoSch92, line 1: I can suggest to test this PR/branch in the OSS VM such that we are sure that the migration works. But probably will works eithouth problems

openhands/automation/git_sync/loop.py
```diff
@@ -501,6 +507,23 @@ async def _validate_and_resolve_fields(
...
+        and automation_state == AutomationState.DRAFT
+        and existing.state != AutomationState.DRAFT
+    ):
+        raise ValueError("existing automations cannot be moved to draft state")
+    expected_enabled = automation_state_enabled(automation_state)
+    if fields.get("state") is not None and fields.get("enabled") is not None:
+        if enabled != expected_enabled:
+            raise ValueError("enabled must be true only when state is ACTIVE")
+    enabled = expected_enabled
+
```
VascoSch92: I think we can write this in a more easy way

```python
raw_state = fields.get("state")
raw_enabled = fields.get("enabled")

requested_enabled = parse_automation_enabled(raw_enabled)
if requested_enabled is None:
    requested_enabled = True

state = model_automation_state(raw_state, requested_enabled)
enabled = automation_state_enabled(state)

if existing is not None and state == AutomationState.DRAFT != existing.state:
    raise ValueError("existing automations cannot be moved to draft state")

if raw_state is not None and raw_enabled is not None and requested_enabled != enabled:
    raise ValueError("enabled must be true only when state is ACTIVE")
```

openhands/automation/git_sync/loop.py
```diff
@@ -501,6 +507,23 @@ async def _validate_and_resolve_fields(
         fields.get("setup_script_path"), "setup_script_path"
     )
     timeout = validate_automation_timeout(fields.get("timeout"))
+
+    enabled = parse_automation_enabled(fields.get("enabled"))
+    if enabled is None:
+        enabled = True
+    automation_state = model_automation_state(fields.get("state"), enabled)
```
VascoSch92: can we call that just `state`?

openhands/automation/utils/state.py
```diff
@@ -0,0 +1,40 @@
...
+        AutomationState.ACTIVE
+        if parse_automation_enabled(enabled)
+        else AutomationState.INACTIVE
+    )
+
+
+def automation_state_enabled(state: AutomationState | str | Enum | None) -> bool:
+    if state is None:
+        return True
+    return AutomationState(_state_value(state)) == AutomationState.ACTIVE
```
VascoSch92:
```suggestion
    return AutomationState(_state_value(state)) == AutomationState.ACTIVE if state else True
```

openhands/automation/utils/state.py
```diff
@@ -0,0 +1,40 @@
...
+def _state_value(state: Any) -> Any:
+    return state.value if isinstance(state, Enum) else state
+
+
+def parse_automation_enabled(enabled: Any) -> bool | None:
+    """Parse the deprecated enabled flag using Pydantic bool coercion."""
+    if enabled is None:
+        return None
+    return _BOOL_ADAPTER.validate_python(enabled)
+
```
VascoSch92:
```suggestion
    return _BOOL_ADAPTER.validate_python(enabled) if enabled is not None else None
```

openhands/automation/models.py
```diff
@@ -48,6 +48,14 @@ class AutomationRunStatus(enum.Enum):
     SKIPPED = "SKIPPED"
 
 
+class AutomationState(enum.Enum):
```
VascoSch92: You can use StrEnum

openhands/automation/schemas.py
```diff
@@ -335,6 +340,44 @@ class RunStatus(StrEnum):
     SKIPPED = "SKIPPED"
 
 
+class AutomationState(StrEnum):
```
VascoSch92: Is that not a repetition? We do have already this class right?

migrations/versions/026_add_state_and_trigger_source.py
malhotra5: Tested this locally with an agent canvas instance to confirm that migrations run as expected.

openhands/automation/git_sync/loop.py
malhotra5: Addressed in e5aba0d by simplifying the git-sync state/enabled handling into raw_state/raw_enabled, requested_enabled, state, and enabled.

openhands/automation/git_sync/loop.py
malhotra5: Addressed in e5aba0d by renaming automation_state to state in the git-sync import validation path.

openhands/automation/models.py
malhotra5: Addressed in e5aba0d by making the model AutomationState a StrEnum.

openhands/automation/schemas.py
malhotra5: Addressed in e5aba0d by removing the duplicate schema AutomationState definition and reusing the model enum.

VascoSch92 (approved):

PR #445, opened 2026-09-11
feat: add user-authenticated KV access

tofarr:
## Why

The KV store previously only supported authentication via per-run KV JWT tokens (`AUTOMATION_KV_TOKEN`). This works for automation scripts executing during a run, but spawned conversations (e.g., the datadog-error-monitor's investigation agent) and users outside the automation run context have no way to access KV state without yet another authentication mechanism. User-authenticated KV access reuses the existing user auth infrastructure (API key / `X-Session-API-Key` / cookie) rather than introducing a new auth method.

## Summary

- Added `KVAuthContext` dataclass holding `automation_id` + `auth_method` (`"kv_token"` | `"user"`)
- Added `_try_resolve_user` dependency that wraps `authenticate_request`, catching 401 and returning `None` so the KV token path can proceed without user credentials (only 401 is swallowed; 429/502 propagate)
- `get_kv_auth_context` checks KV JWT token first (backward compatible), then falls back to user auth with `automation_id` query parameter + org membership verification
- All KV endpoints updated from `Depends(get_token_claims)` → `Depends(get_kv_auth_context)`
- Added 14 tests covering: user auth CRUD, missing `automation_id`, wrong org, nonexistent automation, `X-Session-API-Key` header, KV token backward compat, and auth method detection (API key vs JWT, invalid JWT, no auth)

## Screenshots

image
image

## How to Test

```bash
cd /projects/automation
python3 -m pytest tests/test_kv_user_auth.py -v
```

All 14 tests pass. The existing `tests/test_kv_router.py` suite also passes with the updated dependency override.

## Notes

- KV JWT token auth is fully backward compatible — existing automation scripts using `AUTOMATION_KV_TOKEN` continue to work unchanged
- User auth requires the `automation_id` query parameter (e.g., `GET /v1/kv/state?automation_id=<uuid>`) and verifies the user belongs to the automation's org
- `_try_resolve_user` only catches 401 from `authenticate_request`; other errors (rate limiting, server errors) propagate to the client

---


hieptl (approved):
Thank you! 🙏

PR #446, opened 2026-09-12
feat(slack): continue replies in owned threads

### LINKED ISSUE #442: Respond to replies to the bot's comments, not only @-mentions
## Summary

The Slack stream source (`openhands/automation/streams/slack.py`) only forwards `app_mention` events. Replies to the bot's *own* comments in a thread are delivered by Slack as plain `message` events (not `app_mention`), so the automation never sees them and cannot continue the conversation from a threaded follow-up. We should also respond to replies to the bot's comments, not only to direct @-mentions.

### Desired Behavior

When a user replies in a Slack thread to a message the bot itself posted, the automation should receive that reply as an event and be able to continue the existing conversation — without the user needing to @-mention the bot again.

Concretely:

- A threaded reply whose `parent_user_id` is the bot's own `user_id` should be forwarded to `accept_event()` with `event_key="message"`.
- A bare (non-threaded) channel message should still be ignored.
- A threaded reply to *another user's* message (`parent_user_id != bot_user_id`) should still be ignored.
- The bot's own messages must never re-trigger the automation.
- Existing `app_mention`-only automations must behave exactly as they do today.

## Background

Slack sends different event types depending on how a message is addressed:

- `app_mention` — a message that @-mentions the bot.
- `message` (`message.channels`) — every channel message, including threaded replies. A reply to the bot's comment carries:
  - `thread_ts` — the thread root timestamp (set for any threaded reply).
  - `parent_user_id` — the author of the message being replied to (the bot's own `user_id` when replying to a bot comment).

Today the stream drops everything that is not `app_mention`:

```python
SUPPORTED_EVENT_TYPES = frozenset({"app_mention"})
```

```python
        event = envelope.get("event") or {}
        event_key = event.get("type")
        if event_key not in SUPPORTED_EVENT_TYPES:
            return None
```

So a human replying to the bot's answer in a thread produces a `message` event that is acked and discarded, and the bot can only be re-engaged by @-mentioning it again.

## Proposed change

Accept `message` events, but only the subset that is a threaded reply to the bot's own message. Bare channel messages (no `thread_ts`) and replies to other users (`parent_user_id != bot_user_id`) must still be ignored so the bot does not fire on every message in the channel.

```python
SUPPORTED_EVENT_TYPES = frozenset({"app_mention", "message"})
```

In `SlackStreamProvider.accepted_event`, after the existing bot-message guard:

```python
        # Drop the bot's own messages, or an @-mention it writes triggers the
        # automation that wrote it.
        if event.get("bot_id") or event.get("subtype") == "bot_message":
            return None

        # Only thread replies addressed to the bot's own message are of
        # interest: a bare `message` covers every channel post.
        if event_key == "message":
            is_thread_reply = bool(event.get("thread_ts"))
            replies_to_bot = event.get("parent_user_id") == self.bot_user_id
            if not (is_thread_reply and replies_to_bot):
                return None
```

## Implementation notes / suggestions

1. **Event subscription (Slack App config).** The Slack app must be subscribed to the `message.channels` event, which requires the `channels:history` bot scope. Consider `message.groups`/`message.im` too if the bot is used in private channels or DMs. The events need no extra Socket Mode code — they arrive on the same socket.

2. **Keep the existing bot-message drop.** `bot_id` / `subtype == "bot_message"` already prevents the bot from re-triggering itself; the `message` branch must remain after that guard.

3. **`parent_user_id` is the key field.** It is the author of the parent message. Checking `parent_user_id == self.bot_user_id` precisely selects "reply to the bot's comment" without a fuzzy text match or re-mention requirement.

4. **Do not treat `message` as `app_mention`.** Emit `event_key="message"` so trigger filters can distinguish the two. Automations that want both will configure `"on": ["app_mention", "message"]` (or add a `filter`), keeping backward compatibility for existing `app_mention`-only automations.

5. **Downstream script update.** The deployed Slack automation script (`slack-channel-monitor`) currently early-returns on `ev.get("type") != "app_mention"`, so it must also accept the `message` event and skip the "trigger phrase" requirement for these replies (the reply need not @-mention the bot). This lives in the extensions repo, not this one.

6. **Dedup is already covered.** `provider_event_id=envelope.get("event_id")` still dedupes redelivered `message` envelopes.

### Acceptance Criteria

- [ ] A threaded reply to the bot's own comment reaches `accept_event()` with `event_key="message"`.
- [ ] A bare (non-threaded) channel message is still ignored.
- [ ] A threaded reply to *another user's* message (`parent_user_id != bot_user_id`) is still ignored.
- [ ] The bot's own messages never re-trigger the automation.
- [ ] Existing `app_mention` automations behave exactly as today.

This issue was created by an AI agent (OpenHands) on behalf of the user.

palrohitg:
## Why

Slack Socket Mode forwarded `app_mention` events, but an ordinary human reply in an existing Slack thread arrives as a `message` event. That prevented a user from naturally continuing an automation conversation without mentioning the bot again.

The issue's initial `parent_user_id == bot_user_id` proposal does not cover the normal conversational shape because Slack threads are flat: `parent_user_id` identifies the thread-root author. A human mention therefore creates a human-authored root even after the bot replies inside it.

## What changed

- Accept human-authored Slack `message` events when `thread_ts` is present.
- Continue human-rooted replies only when the automation already owns the resolved thread subject.
- Represent that boundary with `AcceptedEvent.existing_subject_only`, defaulting to `false` for existing transports.
- Stop before run creation when a follow-up-only event has no matching owner, including `dispatch_run` triggers.
- Preserve the explicit bot-rooted case from #442: it may continue an existing subject or create a run.
- Continue dropping bare channel messages, bot-authored messages, and unrelated human-rooted threads.
- Leave existing `app_mention`-only automations unchanged.

## Routing decision

| Slack event | Result |
|---|---|
| `app_mention` | Existing behavior: continue the subject or create a run |
| Threaded `message`, bot-authored root | Continue or create a run |
| Threaded `message`, human-authored root | Continue only if this automation already owns the thread subject |
| Threaded `message` in an unrelated human thread | Deduplicated, but no run is created |
| Bare channel message or bot-authored message | Dropped by the Slack provider |

The ownership check uses the existing durable `AutomationRun.subject_key`. It works across restarts and multiple Socket Mode replicas without a process-local thread cache, Slack history lookup, or schema migration.

An automation that wants natural follow-ups must listen for both event keys and resolve the opener and replies to the same subject:

```json
{
  "type": "event",
  "source": "slack",
  "on": ["app_mention", "message"],
  "destination": "continue_conversation",
  "subject_key_expr": "join('/', [team_id, event.channel, event.thread_ts || event.ts])"
}
```

## Verification

Regression coverage includes:

- acknowledgement before routing for human-rooted and bot-rooted replies;
- provider classification of human-rooted replies as follow-up-only;
- mention → owned human-rooted reply continuing the same conversation;
- an unrelated human-rooted thread creating no run;
- a `dispatch_run` trigger being unable to bypass the follow-up-only guard;
- the new flag defaulting off for existing callers;
- bare messages and bot-authored messages remaining ignored.

Results:

- `uv run pre-commit run --files ... --show-diff-on-failure` — passed (Ruff format/lint, pycodestyle, Pyright).
- Non-Docker Slack provider suite — 25 passed, 5 deselected.
- In-memory SQLite routing flow — passed.
- [Real Slack Socket Mode proof](https://github.com/OpenHands/automation/pull/446#issuecomment-5652255962) — human root mention → bot thread reply → human follow-up without another mention; the follow-up reused the owned conversation and finished with `total runs: 1`.
- PostgreSQL-backed tests could not start locally because this environment cannot access its Docker socket; the test bodies were not entered.
- Repository CI workflows are currently awaiting maintainer approval for this fork PR.

The live proof exercises the real Slack connection, `SlackStreamProvider`, `accept_event()`, durable subject lookup, and queued-run coalescing. It intentionally does not run an LLM or Agent Server; it verifies the routing responsibility owned by this repository.

HUMAN: I ran the real Slack mention and unmentioned follow-up flow and reviewed the token-free PASS output.

## Operational scope

The Slack app must subscribe to `message.channels` and have the `channels:history` bot scope for public-channel replies.

The issue also mentions an event-driven `slack-channel-monitor` change in `OpenHands/extensions`. Current `main` there contains a cron-polling implementation and does not contain the referenced `event_payload` / `app_mention` guard, so this PR does not modify that separate implementation.

All temporary `.pr/` review artifacts were removed in `094665a` and will not be merged.

Fixes #442


palrohitg:
@VascoSch92 @enyst, could one of you take a look when you have a moment? This is a focused follow-up to the Socket Mode work from #384: existing `app_mention` behavior stays unchanged, and `message` events are accepted only for replies to the bot’s thread. Fork CI is currently waiting for approval.

palrohitg:
The failed unit-test job is a shared CI dependency failure, not a Slack stream assertion failure: 1,703 tests passed, then all 17 S3 integration cases failed in fixture setup because Docker Hub no longer serves the pinned MinIO image. The registry repair is isolated in #447.

VascoSch92:
The implementation matches the spec in #442 exactly, but I think the spec itself is wrong.

Slack threads are flat, and `parent_user_id` is the thread root's author on every reply, not the author of the message being replied to. The conversations.replies example (https://docs.slack.dev/reference/methods/conversations.replies/) shows every reply carrying the root author's id.

So `parent_user_id == bot_user_id` means "the bot wrote the thread root", not "the user is replying to the bot". In the usual pattern, a human mentions the bot at top level and the automation answers in that thread, so the root stays the human's message and every follow-up is dropped. Running the shapes through accepted_event on this branch:

```text
human top-level '@bot do X'  (app_mention)   -> ACCEPTED
human follow-up in that thread               -> DROPPED   <-- the case #442 wants
human reply under a bot-rooted top-level post -> ACCEPTED
```

The filter only fires when the bot posts a new top-level message, which is the digest case, not the conversational one. It should accept threads the bot has posted into, not only threads the bot rooted.

Could you please double check that? And also see if you can create an automation where you show the behaviour?

palrohitg:
@VascoSch92 Thanks for catching the `parent_user_id` issue. I updated the routing to use durable subject ownership rather than assuming the bot authored the thread root.

I also ran the corrected flow against a real Slack workspace over Socket Mode:

1. a human mentioned the bot at the thread root;
2. the bot replied in that human-rooted thread;
3. the human followed up without another `@mention`;
4. the follow-up continued the same subject-owned conversation, with `total runs: 1`.

The harness exercised the real `SlackStreamProvider`, `accept_event()`, subject ownership lookup, and queued-run coalescing. It intentionally did not start an LLM or Agent Server; the terminal `PASS` is the routing assertion.

### Slack thread

Human-rooted Slack thread with an unmentioned follow-up

### Token-free terminal result

Live routing harness PASS with one run

PR #456, opened 2026-09-13
fix: repair subject release tracking on existing revision 022 databases

### LINKED ISSUE #455: Repair databases stamped022 before subject release tracking was added
### Bug Description

`dev-oss-automation` has an existing database stamped at Alembic revision `025`, but `automation_runs` lacks `subject_released_at`. The database originally applied the earlier form of revision `022`, which added only `subject_key` and the older subject index. Revision `022` was later changed in place before merging to `main`, so both schemas report the same revision and normal upgrades skip the missing DDL.

After upgrading the service to Automation 1.13.1, ORM queries against `AutomationRun` fail with `sqlite3.OperationalError: no such column: automation_runs.subject_released_at`. Agent Canvas consequently returns HTTP 500 when loading conversations.

The earlier `022` schema was not in a tagged release, but it remains in a deployed OSS database. Recreating that database would discard automation and run history, so the supported upgrade path needs a forward repair.

### Expected Behavior

A new migration after revision `025` repairs the older schema through the normal Alembic startup path. It must also be safe for databases created from tagged releases, where the column and corrected index already exist.

### Acceptance Criteria

- [ ] A database stamped `025` whose `automation_runs` table has `subject_key` but lacks `subject_released_at` upgrades normally.
- [ ] Existing automation and run rows are preserved.
- [ ] The live-subject index includes only non-null subjects whose `subject_released_at` is null.
- [ ] A current 1.13.1 database and a fresh database upgrade without duplicate-column or duplicate-index errors.
- [ ] Automated migration tests reproduce the historical schema and cover repaired, current, and fresh upgrade paths.

neubig:
HUMAN:

The failure was reproduced on the deployed `dev-oss-automation` database after upgrading to Automation 1.13.1: the database was stamped `025` but lacked `automation_runs.subject_released_at`, causing HTTP 500 responses when Canvas loaded conversations.

AGENT:

## Why

An OSS Automation database applied an earlier form of revision `022` that added `subject_key` but not `subject_released_at`. The migration was changed in place before it merged to `main`; Alembic therefore sees the deployed database's later revision stamp and never revisits the missing DDL. Current `AutomationRun` queries then fail because the ORM expects the column.

Changing revision `022` again would rewrite migration history already included in tagged releases. This PR instead adds one forward compatibility migration after current revision `025`.

## Summary

- Add revision `026`, which detects the historical schema, adds `subject_released_at`, and replaces the subject index with the current live-subject predicate.
- Leave current databases unchanged when revision `022` already created the complete schema.
- Exercise historical, current, and fresh SQLite migration paths while preserving existing run data and release timestamps.

## Issue Number

Fixes #455

## How to Test

`uv run pytest -q tests/test_subject_release_migration.py`

The regression test first demonstrates that a current ORM query fails against the historical schema, then verifies normal upgrade to head, repeated startup upgrade, downgrade/re-upgrade behavior, preserved rows, and the corrected partial index.

## Notes

This PR contains one migration and its regression test. It does not modify a live database or release Automation.


neubig:
Closing because the malformed revision 022 existed only on the unmerged `vasco/external-conversations` feature branch and was never included in a release. The first revision 022 merged to `main` already contained the complete schema, so no supported release upgrade requires this permanent repair migration.

neubig:
The live failure was real, but the affected schema came from the unreleased `vasco/external-conversations` commit pinned on the OSS dev host. The first tagged release containing revision 022 (1.11.0) already has the complete column and index, so released upgrade paths do not need a permanent repair migration. The dev host database was backed up and repaired operationally.

PR #460, opened 2026-09-14
fix(presets): install SDK into the run virtualenv

### LINKED ISSUE #338: setup.sh installs the SDK outside the run venv when UV_PYTHON is set — silent success, then ModuleNotFoundError in main.py

palrohitg:
## Why

Preset setup creates a run-local `.venv`, but `uv pip install` also honors an inherited `UV_PYTHON`. In Agent Server environments that variable can point at the server interpreter, so the SDK is installed outside the run virtualenv and the generated entrypoint later fails to import it.

The previous `>=3.12` request could also select a free-threaded CPython 3.14 interpreter, for which native SDK dependencies may not have compatible wheels.

## What changed

- Request regular CPython in the supported `>=3.12,<3.14` range for both prompt and plugin presets.
- Resolve the interpreter created inside `.venv` on Unix or Windows.
- Pass that interpreter explicitly to `uv pip install --python`, so inherited `UV_PYTHON` cannot redirect the installation.
- Fail setup immediately if the virtualenv interpreter is missing.
- Verify `openhands.sdk`, `openhands.tools`, and `openhands.workspace` are importable from the run virtualenv before setup succeeds.

## Regression coverage

The executable regression test runs both real setup scripts with a fake ambient `UV_PYTHON=/ambient/agent-server/python`. It covers both virtualenv layouts:

- `.venv/bin/python`
- `.venv/Scripts/python.exe`

For each prompt/plugin and Unix/Windows combination, it verifies the constrained CPython request, the explicit install target, no mutation of the ambient interpreter, and the final import check.

## Verification

- Focused preset suite: 148 passed.
- Full CI-equivalent suite with Docker-backed tests: 1,744 passed, 7 skipped.
- Coverage: 76.07% (required: 76%).
- Ruff format/lint, pycodestyle, Pyright, shell syntax, and prompt/plugin script parity: passed.

HUMAN: Requested implementation of the reported preset isolation failure; automated and local verification is documented above.

Fixes #338


VascoSch92 (approved):
LGTM

thanks

PR #462, opened 2026-09-14
fix: tolerate transient SQLite write contention

### LINKED ISSUE #461: SQLite: tolerate transient write contention in local deployments
### Actual Behavior

Run the local SQLite service with `uv run uvicorn openhands.automation.app:app`, create several scheduled runs, and update automation definitions through the Canvas while dispatcher bookkeeping is writing. A short concurrent write can raise `sqlite3.OperationalError: database is locked` while persisting `conversation_id`. The entrypoint has already started, but the run remains `RUNNING` without its conversation, command, or sandbox identifiers and its Docker runtime requires manual recovery.

SQLite is the documented local/self-hosted database, so ordinary concurrent scheduler, dispatcher, watchdog, callback, and API writes should tolerate short-lived lock contention.

### Acceptance Criteria

- [ ] SQLite connections wait through a bounded period of transient write contention instead of failing at the default short timeout.
- [ ] SQLite uses WAL mode so normal readers do not unnecessarily block the single local writer.
- [ ] Foreign-key enforcement remains enabled for every SQLite connection.
- [ ] A regression test holds a write lock briefly, releases it, and proves a second service connection completes its write without `database is locked`.
- [ ] PostgreSQL engine configuration is unchanged.

### Live Reproduction

This occurred in the isolated Canvas factory on 2026-09-14 while four five-minute role schedules were being paused and recreated through the UI. The failed run had already started its Docker entrypoint, then failed while persisting `conversation_id`; the subsequent failure-state update encountered the same lock and left the run orphaned.

neubig:
HUMAN:

AGENT:

## Why

Short-lived SQLite write contention can currently abort dispatcher bookkeeping after a Docker entrypoint has started. The run then remains `RUNNING` without the identifiers needed for normal verification and cleanup.

## Summary

Configure every local SQLite connection with a 30-second bounded busy timeout, WAL journaling, and foreign-key enforcement. This reuses SQLAlchemy's connection hook and changes no scheduler, dispatcher, or PostgreSQL path.

A regression holds a real `BEGIN IMMEDIATE` write lock, starts a service write on another connection, releases the lock, and verifies that both writes commit. A separate check verifies the three connection pragmas.

## Issue Number

Closes #461.

## How to Test

- `uv run pytest tests/ -q --ignore=tests/integration` — 1742 passed, 7 skipped
- `uv run ruff check openhands/automation/db.py tests/test_db.py`
- `uv run ruff format --check openhands/automation/db.py tests/test_db.py`

The live failure was reproduced in an isolated Canvas factory while four UI-managed schedules and dispatcher bookkeeping wrote concurrently. The orphaned runtime was released through the SDK after its worker exited.

## Relationship

Independent `main`-based fix. It does not depend on the software-factory dispatch/profile stack.


neubig:
@all-hands-bot Please review the current head `2031d652998f0019c892c350ea40a2f85b819064` and, if there are no blocking findings, explicitly submit an approval rather than only a comment.

PR #490, opened 2026-09-18
fix: release conversation subject in cancel terminal update

### LINKED ISSUE #472: Release conversation subjects when runs are cancelled
### Actual Behavior

In the Docker-backed Canvas deployment, run `e442f366-3a82-497b-9120-90a7738a214e` was canceled after its model connection stalled. A subsequent developer scan submitted the same subject turn and received HTTP 202, but created no replacement run because `subject_released_at` remained unset.

This can be reproduced by canceling a running conversation-turn child with `curl -X POST .../runs/<run-id>/cancel`, then resubmitting its original idempotency key. The submission remains deduplicated against the canceled run.

### Desired Behavior

Cancellation releases subject ownership in the terminal state update, allowing the same idempotent work to be retried in its deterministic conversation after runtime cleanup.

### Acceptance Criteria

- [ ] Canceling a conversation-turn run records `subject_released_at` in the same terminal update.
- [ ] Canceling an ordinary automation run does not claim or release a subject.
- [ ] A later submission for the canceled subject and idempotency key can create a new tracked run in the same deterministic conversation.
- [ ] Focused cancellation and subject-turn regression tests pass.

onatozmenn:
HUMAN:

I ran the targeted pytest suite and confirmed all 109 tests pass, including the two new subject-release cases that fail on base and pass with the fix, with ruff checks also clean.

---

## Why

Cancelling a run that holds a conversation subject only released it when the run had recorded a `sandbox_id` (and in a second write after the terminal update). Local-mode and conversation-turn child runs often have no `sandbox_id`, so `subject_released_at` stayed unset and a resubmitted subject kept deduplicating against the cancelled run instead of creating a replacement (#472).

## Summary

- Record `subject_released_at` atomically in the cancel terminal `UPDATE` (conditional on `subject_key`), independent of `sandbox_id`.
- Drop the now-redundant post-update release; sandbox cleanup path unchanged.
- Add regression tests: subject run without sandbox releases, ordinary runs untouched, cancelled subject routes resubmission to a new run via `continue_conversation`.

## Issue Number

Fixes OpenHands/automation#472

## How to Test

`python -m pytest tests/test_cancel_run.py tests/test_conversations.py tests/test_watchdog.py -q` — 109 passed. The two new subject tests fail on base and pass with the fix. `ruff check` / `ruff format --check` clean.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


onatozmenn:
Good catch — the resubmission test now goes through the real routing path: it cancels a queued subject run, then calls `continue_conversation` for the same subject and asserts `needs_run is True`. On base this fails (the turn gets coalesced into the dead run, so no replacement run is ever created); with the fix it passes.

VascoSch92 (approved):
thanks

